PR #17, opened 2023-01-07
Refactor pushDataToClients

ysard:
The use of buffer allows to reduce the time spent processing characters one by one.
Thus it will lighten main().


danielkucera:
The same here: only one byte per loop

danielkucera:
Closing as obsolete.

PR #19, opened 2023-01-07
Refactor handleNewClient

ysard:
- limit code redundancy

I insert disableTX/enableTX inline functions for easier comprehension of what is done with digitalWrite HIGH/LOW, because the logic is inverted (HIGH to disable).
It will be used in other commits for the whole source.
Also, comments and doc will be added in another PR.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -64,7 +64,7 @@ inline void enableTX() {
 }
 
 void reset(){
-  digitalWrite(TX_DISABLE_PIN, 1);
+  disableTX();
   pinMode(TX_DISABLE_PIN, INPUT_PULLUP);
   ESP.restart();
 }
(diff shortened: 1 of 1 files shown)


danielkucera:
Please resolve conflict.
And one question: are you also testing the changes? With hardware?

ysard:
Done, yes of course all changes are tested on a custom hardware with my ebusd config.
It's not your interface, but it's an ESP12-E quickly soldered on a Chaffoteaux thermostat (ON/OFF but running using eBUS) with a common ground plane and independent 3V3 supply.
I'm using It's RX/TX parts to sniff & send data. Original PCB uses an Atmega369 & reversing it is not so difficult even if I don't understand everything. I will publish that on my repo, there is no variable pot on their hardware, it may be useful.

PR #30, opened 2023-03-23
Track changes from @yuhu-

danielkucera:


yuhu-:
Hi Daniel, maybe you have time and you will take a look at this branch. It's far from perfect, but it serves my personal purposes. Besides the ebus classes under lib, the schedule capability is the main part of the changes. The commands (currently only master-slave) from the commandTable adapted to the heating system (command.hpp) are constantly repeated. The results are also stored there and are delivered as a json object from the web server. The also added message files are currently useless. The code is tested with an adapter of hardware version 4.0

Example entry for commandTable - **Attention - adaptation to your heating system is mandatory**
```
{"08b509030d0600", "Aussentemperatur", "°C", datatype::d2c, 1, 0, 0, 0},
```

include/schedule.hpp
```diff
@@ -0,0 +1,55 @@
...
+    d1b, // DATA1b
+    d1c, // DATA1c
+    d2b, // DATA2b
+    d2c, // DATA2c
+    flt  // float
+};
+
+struct Command
+{
+    const char *data;    // ebus command as ZZ PB SB NN DBx
```
danielkucera: maybe `uint8_t` ?

lib/ebus/Sequence.cpp
```diff
@@ -0,0 +1,234 @@
...
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * ebus is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with ebus. If not, see http://www.gnu.org/licenses/.
```
danielkucera: what does this serve for? can you add a short description to each file?

src/main.cpp
```diff
@@ -217,11 +236,16 @@ char* status_string(){
   pos += sprintf(status + pos, "async mode: %s\n", USE_ASYNCHRONOUS ? "true" : "false");
   pos += sprintf(status + pos, "software serial mode: %s\n", USE_SOFTWARE_SERIAL ? "true" : "false");
   pos += sprintf(status + pos, "uptime: %ld ms\n", millis());
+ // #ifdef ESP32
```
danielkucera: useless?

src/main.cpp
```diff
@@ -217,11 +236,16 @@ char* status_string(){
...
   pos += sprintf(status + pos, "last_connect_time: %ld ms\n", lastConnectTime);
   pos += sprintf(status + pos, "reconnect_count: %d \n", reconnectCount);
+ // #endif
   pos += sprintf(status + pos, "rssi: %d dBm\n", WiFi.RSSI());
   pos += sprintf(status + pos, "free_heap: %d B\n", ESP.getFreeHeap());
   pos += sprintf(status + pos, "reset_code: %d\n", last_reset_code);
+  #ifdef ESP8266
+  pos += sprintf(status + pos, "reset_info: %s\n", ESP.getResetInfo().c_str());
```
danielkucera: What does this offer extra in comparison to `getResetInfoPtr`?

src/main.cpp
```diff
@@ -244,6 +268,61 @@ void handleStatus()
   configServer.send(200, "text/plain", status_string());
 }
 
+void handleJsonStatus()
```
danielkucera: Can you implement this by extending the status_string(bool json)?
My idea would be to replace `sprintf` by `print_val(bool json, char* name, char* unit, ...)`

src/main.cpp
```diff
@@ -307,6 +386,10 @@ void setup() {
   WiFi.onEvent(on_connected, WiFiEvent_t::ARDUINO_EVENT_WIFI_STA_CONNECTED);
 #endif
 
+#ifdef ESP8266
+  WiFi.setAutoReconnect(true);
```
danielkucera: nice!

src/main.cpp
```diff
@@ -382,16 +468,26 @@ void loop() {
 
   wdt_feed();
 
-#ifdef ESP32
+// TODO iotWebConf.doLoop() enabled - is there a reason why it was disabled?
```
danielkucera: I think it introduced quite high latency on esp8266 - you probably don't see this because you are not connecting with ebusd.

src/main.cpp
```diff
@@ -382,16 +468,26 @@ void loop() {
...
+// #ifdef ESP32
   iotWebConf.doLoop();
-#endif
+// #endif
 
   if (WiFi.status() != WL_CONNECTED) {
-    reset();
+    lastConnectTime = 0;
+    // TODO call of reset() commented out in order to prevent restarts in listening mode
```
danielkucera: even in listening mode the adapter is useless when disconnected from network, isn't it?

src/main.cpp
```diff
@@ -382,16 +468,26 @@ void loop() {
...
+  else {
+    if (lastConnectTime == 0) {
+      lastConnectTime = millis();
+      reconnectCount++;
+    }
   }
 
   if (millis() > last_comms + 200*1000 ) {
-    reset();
+    // TODO call of reset() commented out in order to prevent restarts in listening mode
```
danielkucera: maybe condition by running scheduler? or reset the value from scheduler?

danielkucera (commented):
Please see inline comments.
Could you also split this into two PRs where the first will be structural changes to base code and second one will be the implementation of scheduler?

src/main.cpp
yuhu-: human-readable information about the last reset, e.g. ‘Hardware Watchdog’

src/main.cpp
yuhu-: leftovers from the cleaning job

yuhu-:
> Could you also split this into two PRs where the first will be structural changes to base code and second one will be the implementation of scheduler?

I will add a new branch for the base code changes or is there another way?

src/main.cpp
yuhu-: That's what I thought at first, but I needed a way to transmit the json string. However, the web server is still running, but is no longer accessible. Maybe an active sending of the data via http-post is more suitable.

src/main.cpp
yuhu-: in listening mode yes ... but the scheduler collects the data in the meantime too
by the way the comment is wrong - this wasn't the problem with the listening mode

src/main.cpp
yuhu-: This was a real game changer in terms of restarts in listening mode. A connected message client never sent data and last_comms was therefore never updated. There was therefore a reboot every 200 seconds.
=> Maybe this isn't correct anymore. I will test it.

src/main.cpp
yuhu-: `Fatal exception:4 flag:1 (Hardware Watchdog) epc1:0x4020a152 epc2:0x00000000 epc3:0x00000000 excvaddr:0x00000000 depc:0x00000000`
This is a recent example of why I found this information helpful. At random points the adapter restarts with the following message. At the moment I have no idea what triggers this.

src/main.cpp
yuhu-: I have tested the old code and it works. Maybe something went wrong during programming and testing (updating last_comms). I will undo the changes, but without any communication the adapter resets itself every 200 seconds for nothing. I don't think this is necessary behavior.

yuhu-:
- http://esp-ebus.local/json/status provides general information about the system itself, the arbitration and the last command sent
- http://esp-ebus.local/json/statistic gives you some information about the health status of the ebus system. The statistical data is deleted by setting the pwm_value.
- http://esp-ebus.local/json/data provides the collected data of the commands sent.

Just give it a try. By the way, I have removed the reset information because of the pull request https://github.com/danielkucera/esp-arduino-ebus/pull/71.

include/schedule.hpp
yuhu-: It is easier to read and maintain as a character string. And I don't think there is a performance problem with it either.

lib/ebus/Sequence.cpp
yuhu-: I have added a short description to each header file. What is the purpose of a licence in the file? Well, the code has a certain age, and it has already been published under this licence, so it is necessary to add it.

yuhu-:
Quick Info - I have removed the previously implemented json pages. Instead, the data, status and statistics are sent to an mqtt server.

yuhu-:
Basic MQTT support is implemented. 

- MQTT server, user and password can be set on the configuration page. 
**Attention - MQTT data is currently still sent in plain text.** 

- MQTT topics are currently not editable and start with "ebus/". 
- "**ebus/device/**" like the status page, except arbitration. 
- "**ebus/arbitration/**" and “**ebus/messages/**” provide statistical information about the ebus. 
- "**ebus/values/**" holds the collected data of the time-controlled messages sent.
- "**ebus/config/**"
-- Provides the option of inserting, removing and listing installed commands.
-- The commands are read by the MQTT server under the topic "ebus/config/commands/xyz".
-- "**ebus/config/list**" - publish all installed commands under the topic "ebus/config/installed/xyz".
-- "**ebus/config/remove**" - removes the given command e.g. {"command": "08b509030d0600"}
-- "**ebus/config/commands/xyz**" - xyz as topic e.g. 08b509030d0600 
    **payload** {"command":"08b509030d0600","unit":"°C","interval":60,"position":1,"datatype":"DATA2c","topic":"ebus/values/Aussentemperatur"}
-- "**ebus/config/installed/xyz**" - xyz as topic e.g. 08b509030d0600 
    **payload** {"command":"08b509030d0600","unit":"°C","interval":60,"position":1,"datatype":"DATA2c","topic":"ebus/values/Aussentemperatur"}

grafik

yuhu-:
**“ebus/config/remove”** has been removed.
Commands are automatically inserted and removed by inserting/removing them under the topic **"ebus/config/commands"**.

Discovery topics for the Home Assistant are generated if **ha** is set to **true** - for example

**ebus/config/commands/08b509030d0600**
{
  "command": "08b509030d0600",
  "unit": "°C",
  "interval": 60,
  "position": 1,
  "datatype": "DATA2c",
  "topic": "Aussentemperatur",
  "ha": true,
  "ha_class": "temperature"
}

**homeassistant/sensor/ebus/Aussentemperatur/config**
{
 "name": "Aussentemperatur",
  "device_class": "temperature",
  "state_topic": "ebus/values/Aussentemperatur",
  "unit_of_measurement": "°C",
  "unique_id": "08b509030d0600",
  "value_template": "{{value_json.value}}"
}

danielkucera:
Hi @yuhu- ,
does this also work for passive listening without sending any message? Just watch messages and publish values?

yuhu-:
> Hi @yuhu- , does this also work for passive listening without sending any message? Just watch messages and publish values?

I'm not quite sure what you mean. 
- For values, you need a suitable configuration based on the available ebus equipment.
- Messages in raw format can be sent very easily, but that would generate a lot of traffic.
- For passive listening I would add some filters for the really interesting messages
- for e.g. only message to target **08** should be publish on topic xy
2024-11-22 22:35:35.974 - 10**08**b509032907004f00050700930100a900
2024-11-22 22:35:36.156 - 1025b5040101d200092e0300000003000100f100
2024-11-22 22:35:36.276 - ff53b509030d0300e40003d201005b00
2024-11-22 22:35:36.394 - 1025b504020d00be00055000a7032e0400
2024-11-22 22:35:36.583 - 1025b5040132e1000a000000000000e0040000e700
2024-11-22 22:35:36.733 - 1025b5040131e2000200002c00
2024-11-22 22:35:36.893 - 10**08**b5090329b801fd0003b801007000
2024-11-22 22:35:37.742 - 1050b5040100ff000a033835222211052440043700
2024-11-22 22:35:39.212 - ff**08**b509030d0000970003a70300f400
2024-11-22 22:35:39.994 - 10**08**b509032901002300050100fc0100d800
2024-11-22 22:35:40.131 - 00
-----ERROR-DETECTED-----> master sequence is too short
2024-11-22 22:35:40.191 - 10feb5050427002000f7
2024-11-22 22:35:40.274 - 030ab503020001dc000affffffffffffffffffff0f00
2024-11-22 22:35:40.674 - 10feb505034a0100f4

yuhu-:
> Hi @yuhu- , does this also work for passive listening without sending any message? Just watch messages and publish values?

Here you go for passive listening. Just watch and publish interested values...for e.g.
```
{
"command":"fe070009",
"unit":"°C",
"active":false,
"interval":0,
"master":true,
"position":1,
"datatype":"DATA2b",
"topic":"outside_temperature",
"ha":false,
"ha_class":"temperature"
}
```
**ebus/values/outside_temperature {"value":6.4375}**

yuhu-:
Hi, I have added the publication of the raw values. 
**“ebus/config/raw”** with true or false you can start and stop the publication of the raw data. 
**“ebus/config/filter“** as json array for e.g. **[”0700”,...]** only publishes messages where the specified sequence is contained in the master sequence.
The published messages appear under **"ebus/values/raw/"...**
- Topic = Master
- Value = Slave

image

danielkucera:
There are too many commits here which I am getting notifications for. I am closing this as we can discuss in separate PRs.

PR #37, opened 2023-04-13
Implement full ebus arbitration protocol

guido4096:
1) Only start arbitration when the bus is in the correct state 2) Participate in second round arbitration when needed Additional changes
1) Make sure all bus data is send to all clients
2) Add possibility to log messages
3) Minor improvements in code

### DIFF
--- a/include/ebusstate.hpp
+++ b/include/ebusstate.hpp
@@ -1,0 +1,75 @@
+#include "main.hpp"
+
+enum symbols {
+    SYN = 0xAA
+};
+class EBusState {
+public:
+    enum eState {
+        eStartup,                      // In startup mode to analyze bus state
+        eReceivedFirstSYN,             // Received SYN
+        eReceivedAddressAfterSYN,      // Received SYN ADDRESS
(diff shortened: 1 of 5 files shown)


danielkucera:
Thank you very much I see a lot of good stuff here. 
Could you please split it logically into separate PRs so I can review it easier?
- passing output variables as arguments
- espota upload
- arbitration handling
- http response for status
- ....

guido4096:
ok, will do. The espota should actually not have been part of the request. And I'm till doubting for the "http response to status" if it should be included. This is sort of a hack, I'm sure there must be better ways to do logging.

danielkucera:
> ok, will do. The espota should actually not have been part of the request. And I'm till doubting for the "http response to status" if it should be included. This is sort of a hack, I'm sure there must be better ways to do logging.

the browsers I tested open the 5555 port just fine even without http header so I was not bothered to include it, try:
http://esp-ebus.local:5555

guido4096:
on Safari it does need those headers. This is change that is anyways needed. I had also added a debug type messaging into the status output, but that is ugly.

PR #39, opened 2023-04-13
Safari needs header, otherwise complains http/0.9

guido4096:
Safari needs the header HTTP/1.1 200 OK, otherwise it rejects the connection with "Cancelled load ... because it is using http/0.9"

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -134,6 +134,7 @@ bool handleStatusServerRequests() {
   WiFiClient client = statusServer.available();
 
   if (client.availableForWrite() >= AVAILABLE_THRESHOLD) {
+    client.printf("HTTP/1.1 200 OK\r\n\r\n");
     client.printf("uptime: %ld ms\n", millis());
     client.printf("rssi: %d dBm\n", WiFi.RSSI());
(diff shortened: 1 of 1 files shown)


guido4096:
See also here: https://forum.arduino.cc/t/esp8266-responds-using-http-0-9/1005265

danielkucera:
I was thinking about adding a simple command protocol to this port, something like:
- check if there is some data coming from the client together with the new connection
- if there is data
  - if it is an HTTP request - respond with HTTP response header and status
  - if it is a command (not sure about the protocol yes), e.g.: `reset`, `reboot`, `R`, ... - execute the action
- else -  response with status
- close the connection

Or maybe adding a separate port with simple HTTP server and implement all of this there and keep this server simple as is is.

So it has the backward compatibility but adds new features.

guido4096:
adding a command protocol would be nice, will also be useful to trigger ondemand wifi config. 
I don't care myself at this moment for the backward compatibility of the status server... .

Maybe then add a check for GET /status so the header can be added and the raw data will be provided as is right now?

danielkucera:
status server working in v7.0 #64

PR #40, opened 2023-04-13
Support the complete ebus arbitration protocol

guido4096:
1) Only allowed to participate when the bus is in state "Received First Syn" 
2) Participate in second arbitration round when priority class matches


guido4096:
P.S. the code became quite a bit more complex. I didn't find a reasonable way to do it easier. The ebus protocol is interesting for arbitration, you really need to monitor the bus continuously to assess the state and to find out who won.

danielkucera:
Thank you for this. I
 just took the naive approach: send address after received syn but it seems to be more complicated. Would you be willing to support this feature in case someone reports an issue with it? May I tag you for help?

guido4096:
of course; I'm still testing here and keep debugging

danielkucera:
@brozikcz maybe you want to test this one?

guido4096:
After lots of testing, I added an additional change that improved the interaction with ebusd significantly. as much as I can tell, ebusd expects to receive the starting SYN symbol when starting the arbitration.

guido4096:
I’m still having unpredictable behavior. There is at least one bug present with the priority class. Better to wait until i finished all testing.

guido4096:
Had a rough day yesterday evening, ebusd was just not interacting well with the adapter at all. This morning, same code, all works more or less flawless. Either related to my devices on the ebus, or to the wifi. 

Anyways, this state is at this moment working well for me. Also added DEBUG_LOG statements that are removed using a #define

guido4096:
Below is a nice example of the arbitration working. 

- I'm writing from master 0xf3 and another device on the bus is writing with master 0x17 . 
- The master address on the bus is then 0x13; 0xf3 is the same priority class and participates in the second round. 
- The second round is started on AUTO-SYN, about 48 miliseconds after the previous SYN was received. 
- The other device with 0x17 is not participating, since it was not in the same priority class, 
- and now my master 0xf3 is winning. 

CMD_START 0xf3
ARB START 0009 0xf3 9 us
ARB MASTER1    0xf3 148 us
ARB SYMBOL     0x13 4235 us
ARB PARTICIPATE SECOND 0xf3 0x13
ARB SYMBOL     0xaa 48641 us
ARB MASTER2    0xf3 1 us
ARB SYMBOL     0xf3 4028 us
ARB WON2       0xf3 4048 us
ARB SEND WON   0xf3 4069 us

guido4096:
By now I observe only the following problems:

1. Sometimes ebusd is not responsing fast enough to a "WON" status. This is likely wifi overhead. Nothing much we can. Ebusd will retry the arbitration.
2. On my ebus network, some masters don't participate in the second round arbitration, where they should and if they would, they would have won. This means nobody will send data on the bus and AUTOSYN will kick in to recover. Nothing much the code can do, except handle the recovery graceful.
3. Sometimes the symbol received after arbitration is not valid, for example 0x0e; if all participants have correct timing, the symbol should always be following the pattern of a valid master address. This is I think a timing problem with retrieving the SYN from the bus and when to send our master address on the bus.

The third item is the most difficult one to resolve. I think with the Arduino framework it is not possible: 
- The loop does a lot of work, for example with the wifi clients
- In the loop, a symbol is retrieved from the bus. When the symbol is a SYN, typically an arbitration is allowed to start
- However there is no guarantee about the delay between "when the SYN was available in the recv buffer" and "when the code retrieves the SYN". The code can only check that there was not yet another symbol available on the bus; meaning the arbitration can start halfway instead of at the exact beginning of the first bit after SYN is received. 

I believe this problem can only be solved when handling the arbitration in an interrupt, which is not supported by the Arduino framework - to my knowledge.

Fortunately it doesn't happen too often.

PiotrHab:
Hi [guido4096](https://github.com/guido4096)
Just a thought to try out, 
put a Wifi handling on a low priority RTOS task (xTaskCreate).
Should be possible within Arduino framework and does not use interrupts which are time sensitive and might cause MCU reset.

guido4096:
@PiotrHab , Thanks for suggestion. I expect that would cause more occurrences of item 1. To fix the remaining issues, I still think there are two things needed:
1) Put the arbitration handling in an interrupt, so we can fully control the timing. This is something with no dependency to ebusd
2) To fix the issues with wifi, the Enhanced protocol needs to be extended to not just ask for arbitration, but instead to ask for full message to be sent. Dependency to ebusd

danielkucera:
@guido4096 , could something like SoftwareSerial help with 1.?

For 2. I was long asking for this but without any reply:
https://github.com/john30/ebusd/issues/857
https://github.com/john30/ebusd/issues/124

guido4096:
> @guido4096 , could something like SoftwareSerial help with 1.?

I don't see how SoftwareSerial can help. What is happening:
1) arbitration is finished on ESP side and the code send "WON" to ebusd. 
2) now ebusd needs to take over and send a character back
The ESP needs to receive the character from ebusd within about 45 ms. If not yet, the AUTOSYN will detect that the bus is idle for 45 ms and will send a SYN.

danielkucera:
Sorry, I meant it as solution for this one:

> Sometimes the symbol received after arbitration is not valid, for example 0x0e; if all participants have correct timing, the symbol should always be following the pattern of a valid master address. This is I think a timing problem with retrieving the SYN from the bus and when to send our master address on the bus.

guido4096:
> For 2. I was long asking for this but without any reply: [john30/ebusd#857](https://github.com/john30/ebusd/issues/857) [john30/ebusd#124](https://github.com/john30/ebusd/issues/124)

ebusd looks like it has grown over time and  It has now many different responsabilities mixed in one code base that is difficult to understand. Responsibilities in one code base that don't belong together. Maybe better to just create something new, without all the overhead of ebusd. Something that is only able to understand the lowest layer of the protocol and leaves all the interpretation to somebody else. The lowest layer should be implemented completely on the ESP. It should forward what it reads - like it already does - and support a full message protocol for sending and including sending the reply of the message to the requester.

guido4096:
> Sorry, I meant it as solution for this one:
> 
> > Sometimes the symbol received after arbitration is not valid, for example 0x0e; if all participants have correct timing, the symbol should always be following the pattern of a valid master address. This is I think a timing problem with retrieving the SYN from the bus and when to send our master address on the bus.

I don't know enough how SoftwareSerial works. Does it allow bit precise interaction with the signal? If you get signaled per bit, at least you know better the exact timing. But still if you don't have interrupts you cannot guarantee that you can reply in the right time.

guido4096:
P.S. with the changes on this branch, I think you are close to what ebusd-esp is doing. I think ebusd-esp will have the same problems with wifi as the code on this branch. Of course if they connect to a dedicated raspberry pi, the problem should not occur as latency is more predictable. But that is just not a right solution, it spreads a responsibility over two different devices. I think the only good solution is what you already requested here https://github.com/john30/ebusd/issues/124. 

Still then, to get everything in shape, the arbitration protocol itself needs to be closer to the hardware than it is now, ... my opinion based on all the testing and timing problems I observed.

danielkucera:
I am definitely interested.

guido4096:
Hi Daniel, have a look at this branch: https://github.com/guido4096/esp8266-arduino-ebus/tree/implement-on-receive-callback

Right now this code is running very stable for me:
1) On the esp32 it uses pre-emption when a byte is received 
2) It equally compiles on the 8266, then does not use pre-emption. It is nicely abstracted away in a few lines of code
3) Got rid of wifi issues by also sending the SYN symbol when arbitration is starting -> puts ebusd in the right state of mind
4) Not seeing any weird arbitration issues

Because of the use of state machine, the code is not as simple as plain lineair code. I don't know a way to make this easier.

guido4096:
There is still code cleanup needed, mostly on naming conventions, so each file gives the same feeling

guido4096:
I'm expecting there is still tweaking possible on how to deal with ebusd; this is the main black box. I do think this code deals well with what is happening on the bus itself.

danielkucera:
Looking good, can you please create a new MR?

guido4096:
Merged the changes into this branch. Looking forward to feedback on how it behaves, further code cleanup needs, input on how to it interacts with ebusd .

include/main.hpp
```diff
@@ -1,22 +1,33 @@
...
+
 bool handleNewClient(WiFiServer &server, WiFiClient clients[]);
-int pushClient(WiFiClient* client, uint8_t B);
+int  pushClient(WiFiClient* client, uint8_t B);
 void handleClient(WiFiClient* client);
 
-int pushEnhClient(WiFiClient* client, uint8_t B);
-void handleEnhClient(WiFiClient* client);
+
+#endif
```
danielkucera, line 37: what is this endif for?

src/main.cpp
```diff
@@ -134,6 +136,33 @@ void setup() {
...
+   message_buffer[DEBUG_LOG_BUFFER_SIZE-1]=0;
+
+   return ret;
+}
+
+#ifdef USE_ASYNCHRONOUS
+#define ASYNC_MODE true
+#else
+#define ASYNC_MODE false
+#endif
```
danielkucera: is this only because of the status server? can it be directly there without defining additional define? or use value 0/1 for USE_ASYNCHRONOUS?

src/main.cpp
```diff
@@ -142,13 +171,21 @@ bool handleStatusServerRequests() {
   WiFiClient client = statusServer.available();
 
   if (client.availableForWrite() >= AVAILABLE_THRESHOLD) {
+    client.printf("HTTP/1.1 200 OK\r\n\r\n");
```
danielkucera: I am still not sure about this one

include/bus.hpp
```diff
@@ -0,0 +1,45 @@
+#ifndef _BUS_H_
+#define _BUS_H_
+#include "main.hpp"
+#include "queue"
+
+// This object retrieves data from the Serial object and
+// let's it flow through the arbitration process.
+// The "read" method will return data with meta information that tells what should be done
+// with the returned data.
+// This object hides if the underlying implementation is synchronous or asynchronous
```
danielkucera: cosmetic: can you please unify line length or not make like breaks?

src/main.cpp
```diff
@@ -134,6 +136,33 @@ void setup() {
...
+   ret = vsnprintf(&message_buffer[message_buffer_position], DEBUG_LOG_BUFFER_SIZE-message_buffer_position, format, aptr);
+   va_end(aptr);
+   message_buffer_position+=ret;
+   if (message_buffer_position >= DEBUG_LOG_BUFFER_SIZE) {
+     message_buffer_position=DEBUG_LOG_BUFFER_SIZE;
+   }
+   message_buffer[DEBUG_LOG_BUFFER_SIZE-1]=0;
+
+   return ret;
+}
```
danielkucera: Can you please move the logging implementation into a separate file?

include/main.hpp
```diff
@@ -1,22 +1,33 @@
...
+#define USE_ASYNCHRONOUS 
 // https://esp32.com/viewtopic.php?t=19788
 #define AVAILABLE_THRESHOLD 0
 #else
 #define AVAILABLE_THRESHOLD 1
 #endif
 
+inline int DEBUG_LOG(const char *format, ...) { return 0;}
+int DEBUG_LOG_IMPL(const char *format, ...);
+//#define DEBUG_LOG DEBUG_LOG_IMPL
```
danielkucera, line 30: Did you notice any performance issues when logging is enabled?
Did you consider using some popular logging libs for Arduino instead of own impl?

src/main.cpp
guido4096: indeed only to include the status. Changed to #define USE_ASYNCHRONOUS 1 and adapted all locations

include/main.hpp
guido4096: to close the include protection started at the top

src/main.cpp
guido4096: I removed it again; I needed the logging to debug, but it can/should stay local for now

src/main.cpp
guido4096: I removed it; I needed the logging to debug, but it can/should stay local for now

include/bus.hpp
guido4096: done

include/main.hpp
guido4096: I did not notice any performance impact, but in production code it should probably be disabled. 
I guess better to use a standard logging api. I have doubts to include and additional dependency, also taking into account what happened with log4j

guido4096: It looks like there is some performance impact because of the logging. Difficult to notice and to be sure, it seems there is more timeouts reported by ebusd when logging is enabled

include/main.hpp
danielkucera: I see now, thanks.

danielkucera (commented):
I am also thinking if all calls to `Serial` object shouldn't be replaced by the `Bus` object including initialization.
What do you think?

guido4096:
> I am also thinking if all calls to `Serial` object shouldn't be replaced by the `Bus` object including initialization.
> What do you think?

maybe/probably yes; don't have a strong opinion....

guido4096:
could you give is a try to see if that indeed helps the code?

guido4096:
actually now implemented to use Bus instead of Serial

guido4096:
the last commit, calculating a delay before sending a byte on the bus, reduces arbitration "restarts" from about 10% to 0% on esp32.

guido4096:
More statistics. Out of 2131 arbitration attempts, 8 had to restart on the first arbitration. And then 92 generated an error, likely on the second round of arbitration. I think this is as good as it gets?

nbr arbitrations: 2131
nbr restarts: 8
nbr lost: 95
nbr won: 1936
nbr errors: 92

guido4096:
well; timing matters. Without tweakers, I get between 5% and 10% errors/restarts during arbitration. See statistics 1 below. With playing with the delay when the send an arbitration symbol, the errors/restarts go to between 1% and 2%. This is really only feasible with the ASYNCRHONOUS method, because only then the timing of receiving the last SYN is a little bit reliable. 

Maybe more tweaking helps. I'm concerned the tweaking depends on the installation and there might not be a setting that works for all. 

// Statistics 1
async mode: true
uptime: 5568101 ms
rssi: -46 dBm
free_heap: 187536 B
reset_code: 3
loop_duration: 800 us
max_loop_duration: 393826 us
version: v5.3-70-g2f4c052-dirty
nbr arbitrations: 31372
nbr restarts: 1940
nbr lost: 1324
nbr won: 28090
nbr errors: 18


// Statistics 2
async mode: true
uptime: 568194 ms
rssi: -43 dBm
free_heap: 188872 B
reset_code: 3
loop_duration: 393 us
max_loop_duration: 126912 us
version: v5.3-70-g2f4c052-dirty
nbr arbitrations: 2814
nbr restarts1: 7
nbr restarts2: 20
nbr lost1: 204
nbr lost2: 2
nbr won1: 2518
nbr won2: 51
nbr errors: 12

guido4096:
getting to 0.5% to 1% restarts/errors with
  #define ARB_FIRST_DELAY_US  -36
  #define ARB_SECOND_DELAY_US  -36
in arbitration.cpp . This is pretty convincing you need to tweak the timing. Maybe the code can do this itself by monitoring and adapting the delay within limits automatically?

async mode: true
uptime: 363890 ms
rssi: -44 dBm
free_heap: 188972 B
reset_code: 3
loop_duration: 132 us
max_loop_duration: 6442 us
version: v5.3-71-gd62c976-dirty
nbr arbitrations: 1651
nbr restarts1: 2
nbr restarts2: 5
nbr lost1: 148
nbr lost2: 0
nbr won1: 1454
nbr won2: 39
nbr errors: 3

guido4096:
@danielkucera , thanks to your hint on SoftwareSerial; I tried@implemented it and it allows to retrieve the exact timing for the start but of the SYN symbol. This helps the code to do 100% accurate timing during arbitration. On 3000 arbitration attempts I got 0 errors/restarts! On esp32, this is the thing that makes the timing really shine. I did not try to make this work on the ESP8266, did not invest how to communicate between threads/interrupts/tasks on that platform, it is not using FreeRTOS.  

There is still sometimes a communication glitch with ebusd.

guido4096:
Hi Daniel, I'm still seeing weird things that I cannot explain with this implementation. I'm still trying to optimize this, but I'm getting mixed feelings on SoftwareSerial. Still debugging. Might also be there is some bug in ebusd. 

When I say "arbitrations", I actually mean "writes to the bus" that may trigger an arbitration, or that may just win from the start.  I run a loop that sends write messages to the bus to stress test the device.

Sending an esp8266 would help to test how that device behaves with this code; but I don't believe the esp8266 can be tweaked much more, it doesn't allow preemptive task switching as the esp32 does. 

P.S. on a side note, the HW 5.1 device I have has a usb c port, but this port is not working, connecting it to my computer doesn't power it. Not a big deal, so far I was able to work around it.

danielkucera:
I'll send you v4.1 the next week.

Did you try to move the jumper from vcc-bus to vcc-usb?

Also to get into USB programing mode, you need to short PROG and GND pins before powering the device.

guido4096:
yes, I did move the jumper, there is no lights and usb is not recognized

danielkucera:
Are you using usb-c - usb-c cable? That doesnt work. You need to use usb-a - usb-c.

guido4096:
ah. indeed using usb-c to usb-c
let met try another

guido4096:
I got a good usb cable, power works, port is visible. But NodeMCU flasher sticks at "Detecting chip type..." Do you have more hints?

guido4096:
never mind, got it, it is now back working

danielkucera:
Hi @guido4096 , was your order number 2748?

guido4096:
Yep

guido4096:
Hi @danielkucera , I received two modules, thanks! I just finished testing the esp8266 - version v4. With the original code I had quite some errors icw ebusd, the new code behaves much better, at least for me, icw SoftwareSerial and plain synchronous mode.

I think the three main high-level improvements in the code are:
1) using SoftwareSerial we have better control over the timing for the arbitration
2) sticking to the ebus spec when you can and cannot participate in the arbitration
3) the pure state machine design avoids a byte long wait in the main loop

I think the item 3) helps to get the wifi connection better up and running. 

Note that the esp8266 has less memory than esp32 and SoftwareSerial uses more memory because it needs to store the timestamp. This means you should only use 512 entries for the receive buffer. This is more than 2 seconds, so this should be more than enough. I think even 256 should be enough.

platformio.ini
```diff
@@ -48,6 +49,9 @@ platform = espressif32
 board = esp32-c3-devkitm-1
 build_flags =
     -DRESET_PIN=20
+lib_deps =
+    https://github.com/tzapu/WiFiManager
+    https://github.com/guido4096/espsoftwareserial.git#add-startbit-timestamp
```
danielkucera, line 54: @guido4096 , did you consider creating pull request to the original library?

guido4096:
I did ask: [https://github.com/plerup/espsoftwareserial/issues/287](https://github.com/plerup/espsoftwareserial/issues/287)

P.S. still testing. The ebus of my heating system crashed a few time with the esp8266, now testing with the original firmware.

danielkucera:
Hi @guido4096 ,

can we merge this? Is it running fine?

danielkucera:
Hi @guido4096 ,

do you have any update?

guido4096:
Hi Daniel,

sorry for the late reply. The code itself is running fine, however i see with the esp8266 that my heating system goes faster into error mode. Like once every two days instead of once every 5 days. I think this means the code uses more power?

This is reduced if i use hardwareserial for the rsp8266, i cant say if it is the same as with the original code. 

Ill upload the current code in the evening to have all settings I use depending on chip type. And then ill start another test with the esp8266 while connected with external power, so see if that indeed fixes the bus. 

Guid

danielkucera:
Thank you for the update. Let's merge this one with what you have and then if you wish, you can optimize for esp8266 (but you don't need to invest much time into this honestly).

I am thinking about putting the bus handling into a separate task (outside of loop() ) and I don't want to refactor everything again after this is merged :)

guido4096:
I just pushed the latest changes, they are small. Shall I return the two adapters that you gave me for testing back to you?

danielkucera:
No, you can keep them, thank you for all that work.

PR #44, opened 2023-04-16
Exclude time for status server in max loop duration

guido4096:
This change gives a better view on the typical max loop duration. Without this change, the time needed to handle the status server request is included in the max loop duration, which makes the max loop duration useless.


danielkucera:
A good one, thanks.

PR #46, opened 2023-04-16
Explicitly set RxFIFOFull to 1

guido4096:
ESP32 in Arduino uses heuristics to sometimes set RxFIFOFull to 1, better to be explicit

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -80,6 +80,8 @@ void setup() {
 #ifdef ESP32
   Serial1.begin(115200, SERIAL_8N1, 8, 10);
   Serial.begin(2400, SERIAL_8N1, 21, 20);
+  // ESP32 in Arduino uses heuristics to sometimes set RxFIFOFull to 1, better to be explicit
+  Serial.setRxFIFOFull(1); 
   last_reset_code = rtc_get_reset_reason(0);
 #elif defined(ESP8266)
   Serial1.begin(115200);


danielkucera:
Can you please explain this one?

guido4096:
You need to receive the bytes as soon as they come in, you don't want to wait for the receive buffer the fill up. This means you need to tell ESP32 to interrupt as soon as there is one byte in the receive buffer. HardwareSerial for Arduino does this when baud rate is 57600 or less, I guess they assume that in low baudrate situations they expect byte-by-byte handling is expected. I don't think you should rely on assumptions, instead better to make it explicit. See details on the code below.

In this file:
$HOME/.platformio/packages/framework-arduinoespressif32/cores/esp32/HardwareSerial.cpp

The following code is present:
    // Set UART FIFO Full depending on the baud rate. 
    // Lower baud rates will force to emulate byte-by-byte reading
    // Higher baud rates will keep IDF default of 120 bytes for FIFO FULL Interrupt
    // It can also be changed by the application at any time 
    if (!_rxFIFOFull) {    // it has not being changed before calling begin()
      //  set a default FIFO Full value for the IDF driver
      uint8_t fifoFull = 1;
      if (baud > 57600 || (_onReceiveCB != NULL && _onReceiveTimeout)) {
        fifoFull = 120;
      }
      uartSetRxFIFOFull(_uart, fifoFull);
      _rxFIFOFull = fifoFull;
    }

danielkucera:
This makes sense, thank you.

PR #71, opened 2024-03-11
updated status reporting

HuffYk:
- refactored status reporting to use String
- added additional information to status page including ESP32 chip info


yuhu-:
It would be nice if this could be accepted.

diabolo86:
It would be nice to have firmware information and a restart button on the webpage.

Additional information from the bus (which also be send via mqtt) on the page would also be nice

PR #85, opened 2024-11-17
Static IP address and minor fixes

yuhu-:
- Configuration of static IP address added.
image

- MDNS support within IotWebConf disabled, as it is handled outside with the default HOSTNAME “esp-eBUS”. (IotWebConf uses the Thing Name instead.)

- The storage of the pwm_value in the preferences object has been removed, as this is also stored via IoTWebConf


danielkucera:
Thank you. Again, can you please separate the style changes from functional?

danielkucera:
Very nice, thank you.

PR #94, opened 2024-12-21
suggestions

yuhu-:


include/bus.hpp
```diff
@@ -59,8 +59,8 @@ class BusType {
 
   // Is there a value available that should be send to a client?
   bool read(data& d);
-  size_t write(uint8_t symbol);
-  int availableForWrite();
+  static size_t write(uint8_t symbol);
+  static int availableForWrite();
```
danielkucera: are you sure about these?
moreover, the implementation is not defined static.

yuhu-: These are just suggestions from cppecheck - we can leave it as it is

danielkucera: these two do not really make sense to me, the rest is fine

PR #95, opened 2024-12-21
ebus library classes

yuhu-:


danielkucera:
now I am thinking... would it make sense to create a separate repo from these and include them in the project as lib? what do you think?

yuhu-:
good idea

PR #96, opened 2024-12-22
ebus store schedule - ready for review

yuhu-:
**store**
Implements the storage of active and passive commands. New commands can be added and removed via mqtt. It provides functions for saving, loading and deleting commands in the nvs memory. Commands stored in the nvs are loaded at startup.

**schedule**
This class sends time-controlled active commands to the ebus. All valid received messages are compared with passively defined commands and evaluated if they match. Raw data (including filter function) can also be output. Individual commands are transmitted and the results reported back. The scan results are also published.

**mqtt** Commands
- ebus/config/restart - Restart the device
- ebus/config/insert/NAME_OF_COMMAND - Insert new command
- ebus/config/remove/NAME_OF_COMMAND - Remove loaded command
- ebus/config/list - Publish loaded commands
- ebus/config/raw - Enable/disable the raw data printout
- ebus/config/filter - Insert raw data filter
- ebus/config/load - Loading saved commands
- ebus/config/save - Saving loaded commands
- ebus/config/wipe - Wiping saved commands
- ebus/config/send - Sending of given ebus command(s) once


danielkucera:
Hi @yuhu- ,
is this ready for review or still WiP? Can you add it to the title?

yuhu-:
Hi Daniel! Do you have any idea when you will have time to review it? Maybe you can try out and test the implemented functions.

src/bus.cpp
danielkucera, line 1: these are all formating changes, please provide as a separate PR

src/main.cpp
```diff
@@ -51,13 +55,24 @@ TaskHandle_t Task1;
 DNSServer dnsServer;
 WebServer configServer(80);
 
-char pwm_value[NUMBER_LEN];
-
```
danielkucera, line 55: do not move if not necessary

src/main.cpp
```diff
@@ -51,13 +55,24 @@ TaskHandle_t Task1;
...
+char ebus_address[NUMBER_LEN];
+static char ebus_address_values[][NUMBER_LEN] = {
+    "00", "10", "30", "70", "F0", "01", "11", "31", "71",
+    "F1", "03", "13", "33", "73", "F3", "07", "17", "37",
+    "77", "F7", "0F", "1F", "3F", "7F", "FF"};
+
+char comand_distance[NUMBER_LEN];
+// char external_bus_request[STRING_LEN];
+#endif
+
```
danielkucera: definitions can remain without ifdef
condition only functions

src/main.cpp
```diff
@@ -67,7 +82,7 @@ IotWebConf iotWebConf(HOSTNAME, &dnsServer, &configServer, "", CONFIG_VERSION);
 iotwebconf::ParameterGroup connGroup =
     iotwebconf::ParameterGroup("conn", "Connection parameters");
 iotwebconf::CheckboxParameter staticIPParam = iotwebconf::CheckboxParameter(
-    "Enable Static IP", "staticIPParam", staticIPValue, STRING_LEN);
+    "Static IP", "staticIPParam", staticIPValue, STRING_LEN);
```
danielkucera: unrelated cosmetic, add to formating changes

src/main.cpp
```diff
@@ -78,10 +93,24 @@ iotwebconf::TextParameter netmaskParam =
                               STRING_LEN, DEFAULT_NETMASK, DEFAULT_NETMASK);
 
 iotwebconf::ParameterGroup ebusGroup =
-    iotwebconf::ParameterGroup("ebus", "EBUS configuration");
+    iotwebconf::ParameterGroup("ebus", "eBUS configuration");
```
danielkucera: unrelated cosmetic, add to formating changes

src/main.cpp
```diff
@@ -457,6 +548,9 @@ void handleRoot() {
        "user-scalable=no\"/>";
   s += "</head><body>";
   s += "<a href='/status'>Adapter status</a><br>";
+#ifdef EBUS_INTERNAL
+  s += "<a href='/commands'>Commands</a><br>";
+#endif
```
danielkucera:
```suggestion
  s += "<a href='/commands'>Commands</a><br>";
```

danielkucera (changes requested):
Please see my review.

danielkucera:
> Is it realistic for more than one device to be used in the same installation, except in development situations?

Interesting question. Theoretically, one could have one automation/mqtt server controlling multiple ebus installations. This could be solved with parametrizing the `ebus/` prefix of mqtt path so user can specify `ebus/home/`, `ebus/summer_house/` or `ebus1/` `ebus2/` etc.

I would maybe start by having it defined as #define or a const string not necessarily a config param.
I leave the decision on you.

src/main.cpp
yuhu-: wouldn't there be a dead link without the callback ?

> configServer.on("/commands", [] { handleCommands(); });

src/main.cpp
yuhu-: pwm_value is below the network configuration on the config page. I would prefer to use the same order in the code.

src/main.cpp
danielkucera: It would but 404 is okay

src/main.cpp
danielkucera: okay

src/main.cpp
danielkucera, line 1: can you somehow deal with the commented portions of code?
it all seems to deal with some `external bus request`

yuhu-: I will remove this approach because 
* I have some problems with the current arbitration timing and  
* the internal bus request works pretty well on my adapter.

danielkucera:
also please rebase to latest master, ci build has been fixed

danielkucera (approved):
From my side it is ready to merge.
Shall I merge?

yuhu-:
I think so too. The problem with more than one active device is still open. At first I thought it would be easy to change, but there are a few pitfalls to solve. If there really is a need, we could still change it in the future.

danielkucera:
Thank you.

danielkucera:
@yuhu- , could you please write a few lines how to start using the feature?
How to config mqtt, how to do a scan, how to prepare commands, etc.?

PR #107, opened 2025-02-10
mqtt interface and system info

yuhu-:
* mqtt interface for insert and remove redesigned
* system time and outdoor temperature added

### DIFF
--- a/src/schedule.cpp
+++ b/src/schedule.cpp
@@ -2,6 +2,8 @@
 
 #include <ArduinoJson.h>
 
+#include <sstream>
+
 #include "bus.hpp"
 #include "mqtt.hpp"
 
@@ -69,18 +71,7 @@ void Schedule::setDistance(const uint8_t distance) {
   distanceCommands = distance * 1000;
 }
 
-void Schedule::publishRaw(const char *payload) {
-  JsonDocument doc;
-  DeserializationError error = deserializeJson(doc, payload);
-
-  if (error) {
-    std::string err = "DeserializationError ";
-    err += error.c_str();
-    mqttClient.publish("ebus/config/error", 0, false, err.c_str());
-  } else {
-    raw = doc.as<bool>();
-  }
-}
+void Schedule::publishRaw(const bool enable) { raw = enable; }
 
 void Schedule::handleFilter(const char *payload) {
   JsonDocument doc;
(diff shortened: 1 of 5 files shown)


src/schedule.cpp
```diff
@@ -276,13 +267,36 @@ void Schedule::processPassive(const std::vector<uint8_t> &master,
...
+    payload += ":";
+    payload += ebus::Sequence::to_string(ebus::Sequence::range(master, 7, 1));
+    mqttClient.publish(topic.c_str(), 0, true, payload.c_str());
+
+    topic = "ebus/system/outdoor";
+    std::ostringstream ostr;
+    ostr << ebus::byte_2_data2b(ebus::Sequence::range(master, 5, 2));
+    ostr << " °C";
+    mqttClient.publish(topic.c_str(), 0, true, ostr.str().c_str());
+  }
```
danielkucera: This is system specific. Can you guarantee that every device ever made is supporting these values?

src/store.cpp
```diff
@@ -340,16 +349,17 @@ void Store::deserializeCommands(const char *payload) {
...
+      tmpDoc["command"] = variant[1];
+      tmpDoc["unit"] = variant[2];
+      tmpDoc["active"] = variant[3];
+      tmpDoc["interval"] = variant[4];
+      tmpDoc["master"] = variant[5];
+      tmpDoc["position"] = variant[6];
+      tmpDoc["datatype"] = variant[7];
+      tmpDoc["topic"] = variant[8];
+      tmpDoc["ha"] = variant[9];
+      tmpDoc["ha_class"] = variant[10];
```
danielkucera, line 362: This is why I suggested the config doc to be including a header.
From the header, you could recognize field positions and/or config doc version to assume some defaults when upgrading.

src/schedule.cpp
yuhu-: stupid question, no human can guarantee this so I will remove this

danielkucera: Well, if there was some ebus standard enforcing this we could have some certainty but I don't know about any. That's how I meant the question.

danielkucera: Okay, sorry, it's part of application layer spec.

src/store.cpp
yuhu-: You are right, we have discussed this behavior. To change this, a separate pull request is worthwhile

PR #110, opened 2025-02-12
ESP32: stop Task1 on OTA update

yuhu-:


danielkucera:
Can you please provide some findings about this?

yuhu-:
When the update starts over the air [ArduinoOTA.handle() inside loop()], the Task1 continues to read data from the ebus. Because of the time-consuming update task, the Task1 causes some unpredictable behaviors on the ebus. To prevent this behavior, it is better to stop Task1. After the update is finished, the device will be restarted anyway.

PR #111, opened 2025-02-14
incorrect use of serial buffer status corrected

yuhu-:
- The inadvertent use of bus.availableForWrite() instead of bus.available() led to delays 
-- incorrect use of serial buffer status corrected
- ebus library simplifications (library updaed)
-- unnecessary if's removed in ebus library
-- set lockCounter inside of resetActive
-- call passiveErrors or activeErrors when an error occurs
- Scan of non-existent addresses led to resets 
-- counter for failed 0704 added

### DIFF
--- a/src/schedule.cpp
+++ b/src/schedule.cpp
@@ -56,7 +56,7 @@ Track<uint32_t> requestsError("ebus/requests/error", 10);
 Schedule schedule;
 
 Schedule::Schedule()
-    : ebusHandler(0xff, &busReadyCallback, &busWriteCallback, &activeCallback,
+    : ebusHandler(0xff, &writeCallback, &readBufferCallback, &activeCallback,
                   &passiveCallback, &reactiveCallback) {
   ebusHandler.setErrorCallback(errorCallback);
 }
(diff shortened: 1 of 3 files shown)


src/schedule.cpp
```diff
@@ -43,6 +43,7 @@ Track<uint32_t> errorsActiveSlaveACK("ebus/errors/active/slaveACK", 10);
 // resets
 Track<uint32_t> resetsTotal("ebus/resets/total", 10);
 Track<uint32_t> resetsPassive00("ebus/resets/passive00", 10);
+Track<uint32_t> resetsPassive0704("ebus/resets/passive0704", 10);
```
danielkucera, line 46: can you please explain this variable? what is 0704? what are resets passive?

yuhu-: > Scan of non-existent addresses led to resets
> -- counter for failed 0704 added

- 0704 counts the failed scan attempts of non-existent addresses.
- resets passive means that the internal FSM of the Ebus library must be reset due to such a telegram.

it seems to me that you are not familiar with the eBUS specification

danielkucera: I am sorry, you are right, I am not very familiar with the application layer

PR #113, opened 2025-02-28
Uniform data object for all commands

yuhu-:
As discussed, I have simplified the internal data processing so that all commands are stored in the same data object.


danielkucera:
thank you

PR #114, opened 2025-02-28
output for "Identification (07h 04h)" removed

yuhu-:
There is no guarantee that every device supports this command, so it is better to remove it.


danielkucera:
@yuhu- , can we make this and the time/temp thing as default configuration instead? so it is there by default but can be deleted in case it doesn't fit the installation...

PR #121, opened 2025-04-26
iotwebconf

yuhu-:
details in the commit messages

### DIFF
--- a/include/main.hpp
+++ b/include/main.hpp
@@ -5,9 +5,6 @@
 
 #define MAX_SRV_CLIENTS 4
 
-// never used - how does this work?
-#define STACK_PROTECTOR 512  // bytes
-
 #ifdef ESP32
 #define UART_TX 20
 #define UART_RX 21
--- a/include/main.hpp
+++ b/include/main.hpp
@@ -4,10 +4,8 @@
 #include <WiFiServer.h>
 
 #define MAX_SRV_CLIENTS 4
-// On ESP8266, maximum 512 icw SoftwareSerial, otherwise you run out of heap
-#define RXBUFFERSIZE 512
 
-#define QUEUE_SIZE 480
+// never used - how does this work?
 #define STACK_PROTECTOR 512  // bytes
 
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -614,25 +612,7 @@ void setup() {
...
-  if (preferences.getBool("firstboot", true)) {
-    preferences.putBool("firstboot", false);
-
-    iotWebConf.init();
-    strncpy(iotWebConf.getApPasswordParameter()->valueBuffer,
-            DEFAULT_APMODE_PASS, IOTWEBCONF_WORD_LEN);
-    strncpy(iotWebConf.getWifiSsidParameter()->valueBuffer, "ebus-test",
-            IOTWEBCONF_WORD_LEN);
-    strncpy(iotWebConf.getWifiPasswordParameter()->valueBuffer, "lectronz",
-            IOTWEBCONF_WORD_LEN);
```
danielkucera, line 626: these default values are required for production testing. they have to stay. adapter has to try to connect to that network after reset.

include/main.hpp
```diff
@@ -4,13 +4,9 @@
 #include <WiFiServer.h>
 
 #define MAX_SRV_CLIENTS 4
-// On ESP8266, maximum 512 icw SoftwareSerial, otherwise you run out of heap
-#define RXBUFFERSIZE 512
 
-#define QUEUE_SIZE 480
+// never used - how does this work?
 #define STACK_PROTECTOR 512  // bytes
```
danielkucera: please remove

danielkucera (changes requested):

src/main.cpp
yuhu-: Can we specify this with compiler flags? The shipped firmware shouldn't include this.

danielkucera: The same firmware used for testing is then shipped to customer.

yuhu-: as you wish

PR #127, opened 2025-05-17
Added commands to the WebUI

ajvdw:
I've added the commands to the WebUI. The compiled binary exceeds the 1M now. So the partitions layout needs change probably.

### DIFF
--- a/include/http.hpp
+++ b/include/http.hpp
@@ -4,17 +4,6 @@
 #include <IotWebConf.h>
 #include "store.hpp"
 
-void handleStatus();
-#ifdef EBUS_INTERNAL
-void handleCommandsList();
-void handleCommandsUpload();
-void handleCommandsDownload();
-void handleCommandsLoad();
-void handleCommandsSave();
-void handleCommandsWipe();
-void handleCommandsInsert();
-void handleValues();
-#endif
 void SetupHttpHandlers();
 
 extern WebServer configServer;
--- a/include/http.hpp
+++ b/include/http.hpp
@@ -1,0 +1,21 @@
+#pragma once
+
+#include <WiFiServer.h>
+#include <IotWebConf.h>
+#include "store.hpp"
+
+void handleStatus();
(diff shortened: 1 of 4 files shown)


danielkucera:
Thanks @ajvdw .
The main.cpp is becoming quite bloated. Would you mind creating a separate file (eg. http.cpp) for http handling?

src/main.cpp
```diff
@@ -477,19 +477,18 @@ char* status_string() {
...
 
-void handleValues() {
-  configServer.send(200, "application/json;charset=utf-8",
-                    store.getValuesJson().c_str());
+  if (client.availableForWrite() >= AVAILABLE_THRESHOLD) {
+    client.print(status_string());
+    client.flush();
+    client.stop();
+  }
+  return true;
```
danielkucera: Did you change the order of handleStatusServerRequests() and publishStatus() ?

The diff looks weird.

ajvdw: I probably did, I moved it to http.cpp but decided to move it back since it was not handled by configServer.

danielkucera: Can you please move it back after publishStatus?

include/http.hpp
```diff
@@ -0,0 +1,21 @@
...
+void handleStatus();
+#ifdef EBUS_INTERNAL
+void handleCommandsList();
+void handleCommandsUpload();
+void handleCommandsDownload();
+void handleCommandsLoad();
+void handleCommandsSave();
+void handleCommandsWipe();
+void handleCommandsInsert();
+void handleValues();
```
danielkucera: please remove definitions which are only used locally

danielkucera (approved):
Perfect. Ready to merge?

src/main.cpp
ajvdw: Now the diff should make sense

ajvdw:
Go for it.

danielkucera:
Merged, thank you.

Issue #131, opened 2025-05-17
Download commands - save file dialog?


yuhu-:
Is it possible to open a save file dialog instead of opening the json document?

yuhu-:
..about "save file dialog"

I think it would be better to have just one /commands endpoint.

The commands page should display the currently installed commands by default.

Buttons for 
- download (various formats?)
- upload
- load
- save
- wipe
- values

should be present.

Instead of the save file dialog, the commands could be saved directly.

@danielkucera and @ajvdw What do you think about it?

ajvdw:
I'm look at more than 40 commands (3 devices, heat pump, boiler, thermostat). So I don't think I would manipulate them from the UI. The typical use case is IMHO: Someone downloads the command set for his/her device uploads it and never will never access the WebUI again. 

What is needed is a  location where the different configurations can be downloaded, comparable to the ebusd configuration files. To get up to speed quickly, a converter for ebusd configurations files to commands would help.

yuhu-:
> I'm look at more than 40 commands (3 devices, heat pump, boiler, thermostat). So I don't think I would manipulate them from the UI. The typical use case is IMHO: Someone downloads the command set for his/her device uploads it and never will never access the WebUI again.

Nevertheless, the manipulation of the commands under an end point could be analogous to the configuration.

PR #132, opened 2025-05-18
Applied Google style guide formatting and fixed /config

ajvdw:
As requested in issue #131 
- Changed formatting
- Fixed the /config handler
- Removed unnecessary include
- Minor fixes

I cannot find an easy fix for the File Dialog. Suggestions are welcome. 
TODO: confirmation for load/save/delete

### DIFF
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -1,4 +1,5 @@
 #include "http.hpp"
+
 #include "main.hpp"
 
 WebServer configServer(80);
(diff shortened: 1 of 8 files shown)


src/http.cpp
```diff
@@ -20,94 +12,90 @@ void handleCommandsList() {
...
 
 void handleCommandsInsert() {
   JsonDocument doc;
-  String body = configServer.arg("plain"); 
- 
+  String body = configServer.arg("plain");
+
   DeserializationError error = deserializeJson(doc, body);
 
   if (error)
```
yuhu-: one minor thing...cpplint complains about braces.
`If an else has a brace on one side, it should have it on both`

src/http.cpp
```diff
@@ -20,94 +12,90 @@ void handleCommandsList() {
...
-    {
+    if (array != nullptr) {
       for (JsonVariant variant : array)
-          store.insertCommand( store.createCommand(variant) );
+        store.insertCommand(store.createCommand(variant));
       configServer.send(200, "text/html", "OK");
-    }
-    else
+    } else
       configServer.send(403, "text/html", "NO COMMANDS");
```
yuhu-, line 100: braces

ajvdw:
fixed:
- braces
- layout stoff
- asking for confirmation for big actions (load, save, wipe, restart) through WebUI

yuhu-:
I'm having an issue with this software version. The problems may have started with https://github.com/danielkucera/esp-arduino-ebus/pull/127, but I'm not entirely sure. The problem is a very short operating time of approximately 3-15 minutes. The reset code is 3, which indicates a software reset. Have you had similar experiences?

yuhu-:
This problem is probably due to the ebus library. #133

ajvdw:
What's holding us back to merge this PR?

danielkucera:
There are unresolved comments. After that I'll do my review.

src/http.cpp
ajvdw: This one should be solved.

src/http.cpp
ajvdw: Brackets should be there now.

ajvdw:
I think I addressed all unresolved comments, if not let me know...

danielkucera (approved):

PR #138, opened 2025-06-01
ebus endpoints added

yuhu-:
* http://esp-ebus.local/api/v1/GetAdapter - general adapter information
* http://esp-ebus.local/api/v1/GetStatus - actual status data
* http://esp-ebus.local/api/v1/GetCounters - counters of schedule


danielkucera:
@yuhu- , ready for merge?

yuhu-:
yes

ajvdw:
@yuhu- I'm trying to use the 'internal' variant. I'm having trouble getting any data. The scan does work. So the interface works. But I don't receive the outside temperature. I'd like to monitor the ebus communication by sending it to a debug topic, but I'm having trouble finding a good location to tap into the data stream. Any suggestions?

yuhu-:
@ajvdw 
* I assume you have the corresponding command installed. 
* The value will be publish below ```ebus/8406ac/values/outdoor_temperature``` 
* `8406ac = adapt to your unique device ID`

```
payload:
{
  "id": "insert",
  "commands": [
    {
      "key": "01",                       // unique key of command
      "command": "fe070009",             // ebus command as vector of "ZZPBSBNNDBx"
      "unit": "°C",                      // unit of the interested part
      "active": false,                   // active sending of command
      "interval": 0,                     // minimum interval between two commands in seconds
      "master": true,                    // value of interest is in master or slave part
      "position": 1,                     // starting position in the interested part
      "datatype": "DATA2B",              // ebus datatype
      "divider": 1,                      // divider for value conversion
      "digits": 2,                       // deciaml digits of value
      "topic": "outdoor_temperature",    // mqtt topic below "values/"
      "ha": true,                        // home assistant support for auto discovery
      "ha_class": "temperature"          // home assistant device_class
    }
  ]
}
```
Another option is to forward the raw message to your MQTT server
* `ebus/8406ac/request`
```
payload:
{
  "id": "forward",
  "value": true,
  "filters": [
    "fe070009"
  ]
}
```
The answer could look like this
grafik

I don't know if my answer fits your question. Just let me know.

ajvdw:
I found out I mixed up the config files of ebusd.
Now I can capture 10feb5160301230e which should contain the outside temp.

230E -> smallest first -> 0E23 =  3619. Divide by 256 = 14.137 degrees

I use:

  "id": "insert",
  "commands": [
    {
      "key": "01",
      "command": "feb5160301",
      "unit": "°C",
      "active": false,
      "interval": 0,
      "master": true,
      "position": 2,
      "datatype": "DATA2B",
      "divider": 1,
      "digits": 2,
      "topic": "outdoor_temperature",
      "ha": true,
      "ha_class": "temperature"
    }
    
  ]
}

Why do I need to set position to 2 in order to get it to show the proper values?

yuhu-:
You figured it out right.

> Why do I need to set position to 2 in order to get it to show the proper values?

The position depends on the definition of the payload used in the telegram and was made in this case by Vaillant.

When you look at the raw data and the definition of the broadcast telegram, it becomes clearer...

- Definition: QQ ZZ PB SB NN DBx CRC
DBx – x starts with 1 instead of 0

- RAW: (output of ebusread / not commited yet)

2025-09-19 18:21:11.382 10feb5160301**7014**56
---SERVICE-DETECTED---> feb5160301: 20.4375 °C
```
QQ=10
ZZ=fe
PB=b5
SB=16
NN=03
DB1=01
DB2=70
DB3=14
CRC=56
```

- DB1 appears to be the register address.
- DB2 / DB3 are the corresponding data of this register.
- To obtain the correct data, you need to know the starting position (DBx=2) and the data type (DATA2B) for the length.

> What are typical values for
> Bus ISR window (micro seconds)
> Bus ISR offset (micro seconds) 
```
busisr_window: 4300 us
busisr_offset: 70 us
```
You can check your resulting values ​​at http://esp-ebus.local/api/v1/GetTiming under BusIsr->Window

image

> And Bus Address?

The bus address depends on your participants. Take a look at http://esp-ebus.local/api/v1/GetCounter for the used addresses and chose a free one.

image

ajvdw:
Thanks, I am still having a hard time translating the ebusd configs to commands.
For example in https://github.com/john30/ebusd-configuration/blob/master/ebusd-2.1.x/en/vaillant/15.ctlv2.csv there is a z1RoomTemp. With some data capture using forwarding, I think I managed to come up with 
` 
 {
      "key": "02",
      "command": "15b52406020003000f00",
      "unit": "°C",
      "active": true,
      "interval": 5,
      "master": false,
      "position": 5,
      "datatype": "UINT32",
      "divider": 1,
      "digits": 2,
      "topic": "room_temperature",
      "ha": true,
      "ha_class": "temperature"
}    
`
But is there a way to predict what the position should be? 
The datatype is for debugging anyway since 'EXP' is not supported yet (4 byte float). I would help if there was a datatype 'HEX' that stores the answer in a hex string in the value starting at position.

PR #141, opened 2025-06-29
EBUS_INTERNAL only

yuhu-:
Over the past few weeks, I've revised the eBUS library and fixed timing and logical issues.

The changes resulted in a modified structure and task distribution, which is (currently) only supported with freeRTOS (ESP32). 

    I decided to use only HardwareSerial for serial access.
    Bus access is controlled via ISR and delivers the data to a byteQueue. 
    A dedicated serviceRunner processes this in the handler (FSM). 
    Any output is processed via eventQueue in the waiting scheduleTask.

For this purpose, I've added two new profiles as follows:

    [env:esp32-c3-internal]
    [env:esp32-c3-internal-ota]

EBUS_INTERNAL now truly means internal only. 

    Read access to port 3334 is supported. 
    Write access via port 3333 and port 3335 is disabled. 
    Status queries via port 5555 are also possible.
    The other profiles are now as before (without schedule, store,...but with basic mqtt support)

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,7 @@ Available MQTT commands.
 |participants     |Publishing scanned ebus participants                          |x
 |send             |Sending ebus commands once                                    |x
 |forward          |Activate/deactivate data forwarding (including filtering)     |x
+|reset            |Resetting counter and timing values                           |x
 
 
 ### Examples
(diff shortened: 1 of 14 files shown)


yuhu-:
Is there any chance to accept this pull request?

danielkucera:
Hi @yuhu- , why closed?

yuhu-:
Because it seemed to me that you didn't agree with this approach.

PR #143, opened 2025-08-13
Redesign

yuhu-:
coming soon


include/main.hpp
```diff
@@ -24,11 +26,9 @@
 inline int DEBUG_LOG(const char *format, ...) { return 0; }
 int DEBUG_LOG_IMPL(const char *format, ...);
 // #define DEBUG_LOG DEBUG_LOG_IMPL
+// #define DEBUG_LOG(...) Serial.printf(__VA_ARGS__)
```
danielkucera, line 29: Do these lines have any function?

include/schedule.hpp
```diff
@@ -18,88 +19,136 @@
...
+  bool isValid() const { return vec_070400.size() > 0; }
+
+  bool isVaillant() const {
+    return (vec_070400.size() > 1 && vec_070400[1] == VENDOR_VAILLANT);
+  }
+
+  bool isVaillantValid() const {
+    return (vec_b5090124.size() > 0 && vec_b5090125.size() > 0 &&
+            vec_b5090126.size() > 0 && vec_b5090127.size() > 0);
+  }
```
danielkucera, line 39: Where are these vaillant functions being used?

platformio.ini
```diff
@@ -87,3 +85,32 @@ upload_protocol = espota
...
+    -DARDUINO_USB_CDC_ON_BOOT=1
+    -DBusSer=Serial1
+    -DDebugSer=Serial
+    -DSTATUS_LED_PIN=3
+    -DEBUS_INTERNAL=1
+
+monitor_filters = esp32_exception_decoder
+lib_deps =
+    ${env.lib_deps}
+    https://github.com/yuhu-/ebus#5a5d6bf
```
danielkucera, line 107: Would it be possible to make this as an extension (`extends`) to `esp32-c3`?

src/http.cpp
```diff
@@ -119,24 +115,40 @@ void handleValues() {
...
+  configServer.send(200, "text/html", "Full scan initiated");
+}
+
+void handleScanVendor() {
+  schedule.handleScanVendor();
+  configServer.send(200, "text/html", "Vendor scan initiated");
 }
 
-void handleParicipantsList() {
+void handleParticipants() {
```
danielkucera, line 133: I would prefer then `handleGetParticipants`

include/main.hpp
yuhu-: the first line is old
the second line I used for debugging

src/http.cpp
yuhu-: me too

include/schedule.hpp
yuhu-: Schedule::getParticipantJson()
Schedule::handleScanVendor()

platformio.ini
yuhu-: Yes that is possible, but the internal profile does not use the espsoftwareserial library
If I understood you correctly, you only want to have a profile with focus on esp32-c3 anyway

yuhu-:
> There are only two options now:
> 
> - I will accept because I value your work and will have to trust nothing breaks terribly
> - You rewrite it into smaller logically separated PRs with meaningful name which would be reviewable
> 
> What do you think?

It's up to you what you prefer...

yuhu-:
Hi @danielkucera – what do you prefer: separate PRs or this PR?

danielkucera:
> Hi @danielkucera – what do you prefer: separate PRs or this PR?

I would prefer separate PRs.
Thank you.

platformio.ini
danielkucera: then you should be able to do just:
```
lib_deps =
    ${env.lib_deps}
```
I would prefer all specific profiles derived from one main so it is immediately obvious what is different.

PR #147, opened 2025-10-10
Redesign - move enhanced into client

yuhu-:
- client methods consolidated and moved into client
- wifi client variables and methods names renamed


include/main.hpp
```diff
@@ -5,7 +5,7 @@
 
 #include <string>
 
-#define MAX_SRV_CLIENTS 4
```
danielkucera, line 8: why did you chose to rename this? If we had also ethernet connectivity, wouldn't it be served by the same "SRV" ?
The rest looks okay.

danielkucera: oh, I see, most likely not because this is `WiFiServer`

PR #148, opened 2025-10-10
Redesign - improve command management

yuhu-:
- Store class refactored to use unordered_map for fast lookup


danielkucera:
thank you

PR #151, opened 2025-10-12
Redesign - HTTP related

yuhu-:
- reset endpoint added
- scan endpoints changed
- GetAdapter endpoint merged into GetStatus


danielkucera:
Nice, thank you.

PR #155, opened 2025-10-16
Redesign - internal profiles

yuhu-:
- Adjustments due to a revised eBUS library (timing and logic issues fixed) that only supports ESP32 (freeRTOS).
- Serial access is done exclusively via ESP-IDF UART.
- Bus access is controlled via ISR (Interrupt Service Routine).
- The handling of the clients has been outsourced to a ClientManager.
- The work is divided into separate tasks.
```
Data flow:
- All received data via ISR (ebusUartEventRunner task) is forwarded to a ByteQueue. 
- A dedicated ServiceRunner (task) reads the ByteQueue and forwards the data... 

a) for INTERNAL messages:
- to the handler (FSM).
- The processed data/results are made available to Schedule (task) via an EventQueue.

b) for EXTERNAL messages:
- to the ClientManager (task).
- The data is forwarded byte by byte to the connected clients.
- The eBUS telegrams must be processed externally.
```

- Read access to port 3334 is supported.
- Write access via port 3333 is supported.
- Write access via port 3335 is supported. (ebusd prefix ens: and enh: are possible)
- Status queries via port 5555 are also possible.
- Added query for existence, life signs and automatic scan on startup.
- The other profiles are now as before (without schedule, store,...but with basic mqtt support)

For this purpose, I've added two new profiles for development as follows:
```
    [env:esp32-c3-internal]
    [env:esp32-c3-internal-ota]
```

- The new profiles enable the simultaneous connection of different external client types and the execution of internal commands only with the support of the ebus library.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,8 @@ jobs:
         pip install --upgrade platformio
     - name: Prepare release dir
       run: mkdir release
+    - name: Build v5.x-internal
+      run: pio run -e esp32-c3-internal ; cp .pio/build/esp32-c3-internal/firmware.bin release/firmware-HW_v5.x-internal-${GITHUB_SHA::6}.bin
     - name: Build v5.x
       run: pio run -e esp32-c3 ; cp .pio/build/esp32-c3/firmware.bin release/firmware-HW_v5.x-${GITHUB_SHA::6}.bin
     - name: Assemble v5.x fullflash


yuhu-:
still too big ?

danielkucera:
Looking good. One question though, do we want also build a binary release with internal enabled?

yuhu-:
Sure.

danielkucera:
Can you please adjust `.github/workflows/build.yml` accordingly?

PR #166, opened 2025-11-23
add more HW info into status

danielkucera:


danielkucera:
this will wait until last release for older hardware

src/main.cpp
```diff
@@ -473,6 +473,16 @@ char* status_string() {
                   USE_SOFTWARE_SERIAL ? "true" : "false");
 #endif
   pos += snprintf(status + pos, bufferSize - pos, "unique_id: %s\n", unique_id);
+  pos += snprintf(status + pos, bufferSize - pos, "chip_model: %s\n",
```
yuhu-, line 525: getChipModel and getChipRevision are not available on eps8266

danielkucera: I noticed, that's why this needs to wait for esp8266 support removal.

yuhu-: Is the timeline for removing esp8266 support?
In the mean time you could add it just for ESP32.

danielkucera: I need to test the current firmware with esp8266, make last release and then we can start removing compatibility code.

danielkucera: Few weeks.

yuhu- (approved):

PR #167, opened 2025-11-23
unit_of_measurement added to diagnostic entities

yuhu-:


src/mqttha.cpp
```diff
@@ -250,13 +250,15 @@ MqttHA::Component MqttHA::createDiagnosticUptime() const {
...
   c.fields["state_topic"] = createStateTopic("state", "free_heap");
+  c.fields["unit_of_measurement"] = "B";
   c.fields["value_template"] = "{{value|int}}";
   return c;
 }
 
 MqttHA::Component MqttHA::createDiagnosticLoopDuration() const {
   Component c = createDiagnostic("sensor", "loop_duration", "Loop Duration");
   c.fields["state_topic"] = createStateTopic("state", "loop_duration");
+  c.fields["unit_of_measurement"] = "µs";
```
danielkucera, line 261: does it handle unicode correctly? did you test?

yuhu-: it works in my environment

PR #168, opened 2025-11-23
Update HA device info

danielkucera:


src/mqttha.cpp
```diff
@@ -231,19 +232,21 @@ MqttHA::Component MqttHA::createDiagnostic(const std::string& component,
 }
 
 MqttHA::Component MqttHA::createDiagnosticUptime() const {
+  std::string url = "http://" + std::string(WiFi.localIP().toString().c_str()) + "/";
```
danielkucera, line 235: @yuhu- , I have been thinking how to do this more nicely. 
Can we pass all the device parameters like url, version, etc. from somewhere, where we have them available (main.cpp?) ?

yuhu-: Yeah - include WiFi should not be used in this class

danielkucera: Now when I look at the code, do you think we could instantiate mqttha in main with a constructor with these parameters and then pass it to where necessary (mqtt, http,..)?
I will try...

yuhu-: I prefer set methodes instead

yuhu-: a struct device info in main would serve the purpose (for http, mqtt, status_string,...json)

danielkucera:
closing in favor of #172

PR #169, opened 2025-11-23
mqtt enabled checkbox added + minor code cleanup

yuhu-:


src/main.cpp
```diff
@@ -96,17 +96,17 @@ char busisr_offset[NUMBER_LEN];
 char inquiryOfExistenceValue[STRING_LEN];
 char scanOnStartupValue[STRING_LEN];
 char command_distance[NUMBER_LEN];
-#endif
 
+char mqtt_support[STRING_LEN];
```
danielkucera: I suggest replacing all usages of the word `support` with `enabled/activated`

yuhu-: ok

Issue #176, opened 2025-11-26
Multiple values from an active command


danielkucera:
@yuhu- have solved multiple passive commands from one telegram in #171 

Now I would like to have one active command which will send the telegram and multiple passive that will map the rest of the values.
I find this useful for commands like this one:
```
2025-11-26 20:07:48.355 [update notice] received read bai Status01 QQ=10: 43.0;36.0;-;-;40.0;on
```

Moreover, on my bus, there is a thermostat that usually sends this command. Can we only send the active one, if the interval wasn't met on the bus by another master (the termostat) and treat it as passive when it was?

yuhu-:
> Now I would like to have one active command which will send the telegram and multiple passive that will map the rest of the values.

Active and passive commands are processed completely separately. This means that active commands are not decoded in the same way as passive commands. While your proposed solution achieves its purpose, it also addresses a fundamental design flaw. Although the data types used are simple, commands with multiple values ​​of different types require multiple installed and configured identical commands, which bloats and clutters the entire system.

> Furthermore, there is a thermostat on my bus that typically sends this command. Could we send the active command only when the interval is not met by another master on the bus (the thermostat), and otherwise treat it as passive?

Should the command be sent actively if the expected command is not seen on the bus within a given time window?

danielkucera:
> While your proposed solution achieves its purpose, it also addresses a fundamental design flaw.

Design flaw on which side? How can we address that?

> Should the command be sent actively if the expected command is not seen on the bus within a given time window?

Yes, that is what I meant.

yuhu-:
Your command is a good example. A single command with multiple transmitted measurements and potentially different underlying simple data types. The current command definition requires a separate entry to process each of these data types. That's why I introduced the ebusd CSV files in 2012 and first released them in 2014.

As I mentioned in https://github.com/danielkucera/esp-arduino-ebus/issues/157#issuecomment-3458384545, processing complex data types and/or multiple values ​​in a single command requires a different definition. I hope these examples make that clearer.

@ajvdw What's your opinion?

ajvdw:
> [@ajvdw](https://github.com/ajvdw) What's your opinion?

I think we should not try to solve this in configuration. 
Basically all we want is that certain write_commands are not fired too often.
If they are used by multiple definitions of active commands then the redundant write_commands can be identified (are the same).

[...]

PR #177, opened 2025-11-26
process passive commands for active messages

danielkucera:
Closes: #176


danielkucera:
@yuhu- , do you find this solution appropriate?

yuhu-:
I assume you have already tested this solution, and even without my own tests it fulfills its purpose, but it shadows a fundamental problem with simple data types, the command definition and their processing. Related to [Multiple values from an active command](https://github.com/danielkucera/esp-arduino-ebus/issues/176)

danielkucera:
@yuhu- , are you okay with merging this?

yuhu-:
when it works for you just merge it

PR #180, opened 2025-11-29
Standard field values ​​for serialized data

yuhu-:
@ajvdw and @danielkucera please review this possible solution for [[Bug] Adapter crashes after firmware upgrade](https://github.com/danielkucera/esp-arduino-ebus/issues/179)

### DIFF
--- a/include/store.hpp
+++ b/include/store.hpp
@@ -113,6 +113,7 @@ class Store {
   // Flexible serialization/deserialization
   const std::string serializeCommands() const;
   void deserializeCommands(const char* payload);
+  void getDefaultForField(JsonDocument& doc, const std::string& field);
 };
 
 extern Store store;
--- a/src/store.cpp
+++ b/src/store.cpp
@@ -644,12 +644,42 @@ void Store::deserializeCommands(const char* payload) {
     for (size_t i = 1; i < array.size(); ++i) {
       JsonArray values = array[i];
       JsonDocument tmpDoc;
-      for (size_t j = 0; j < fields.size() && j < values.size(); ++j) {
-        tmpDoc[fields[j]] = values[j];
+      for (size_t j = 0; j < fields.size(); ++j) {
+        if (j < values.size())
+          tmpDoc[fields[j]] = values[j];
+        else
+          getDefaultForField(tmpDoc, fields[j]);
       }
       insertCommand(createCommand(tmpDoc));
     }
   }
+}
+
+void Store::getDefaultForField(JsonDocument& doc, const std::string& field) {
+  if (field == "key") doc[field] = "";
+  if (field == "name") doc[field] = "";
+  if (field == "read_cmd") doc[field] = "";
+  if (field == "write_cmd") doc[field] = "";
+  if (field == "unit") doc[field] = "";
+  if (field == "active") doc[field] = false;
+  if (field == "interval") doc[field] = 60;
+  if (field == "master") doc[field] = false;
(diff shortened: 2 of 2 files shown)


danielkucera:
Can you elaborate a bit why this is needed? How does it solve the crashing?

yuhu-:
Suppose we add a new field to the command structure. This field is also added to the `serializeCommands` fields. Furthermore, suppose that the adapter already contains commands without the new field from an older firmware version. After installing the new firmware, the adapter restarts, and the stored commands are loaded in the old format. Starting with the last field of the header (the new header contains a new field), this data field is missing. This field might not be initialized. The extension ensures that missing fields are initialized with a default value.

ajvdw:
Makes sense. I think it's time to consider if the command structure should be wrapped in a class. In that case you would handle it in the constructor. It also sets us up for storing the commands a little more efficient if we're getting storage issues in NVM.

yuhu-:
>  It also sets us up for storing the commands a little more efficient if we're getting storage issues in NVM.

**more efficient** - how?

danielkucera:
I would suggest a more robust solution: when a command is inserted or loaded from NVS, it would go through a validator. 
If the validation cannot complete, insert reports an error or load from NVS drops the command.

ajvdw:
> > It also sets us up for storing the commands a little more efficient if we're getting storage issues in NVM.
> 
> **more efficient** - how?

Depending on the component type, attributes can be omitted. But a I agree - not worth the effort.

yuhu-:
> I would suggest a more robust solution: when a command is inserted or loaded from NVS, it would go through a validator.
> If the validation cannot complete, insert reports an error or load from NVS drops the command.

You are right. There should be a more robust check of incoming commands, which is currently partially performed in Store::createCommand().

This check will not help you read data stored in the old format using a newer firmware with a changed format.

danielkucera:
> This check will not help you read data stored in the old format using a newer firmware with a changed format.

It could if it was also adding defaults for missing fields as you implemented here.

yuhu-:
Is there a reason why you are waiting to merge?

danielkucera:
I think we were discussing other possiible solutions.
Besides, are you not only iterating over fields that are already present in saved commands?
What about those, that were added in potential firmware upgrade?

yuhu-:
You're right. It's not as simple as it looks. Potential pitfalls with different definitions:

* A field was added.
* A field was removed.
* A field was renamed.

I think about a version number as the first byte. Every definition change results in a new version number. Reading the version byte and selecting the correct definition could be a solution.

By the way, last weekend I had a strange problem: Saving data to the NVS failed. Wiping the data via `preferences.remove("ebus")` worked, but didn't solve the problem. To work around this (possibly a bug), I temporarily flashed a firmware to completely delete the NVS using `preferences.remove()`.

danielkucera:
It's good to make a full dump in case of such problems so we can investigate later.
Btw, if you need some extra adapter for experimenting I can send you.

For the solution, I think we need to save code memory, the firmware is getting quite big already so I wouldn't go for a "solve every possible situation" solution. 

As I said, I would go for a validator that will validate all parameters required for that particular version and fill in missing values with defaults. 
If the previous config cannot be validated, it will simply not load problematic commands.
We could also provide a way to download the saved commands in raw format and then provide an external (web) tool for conversion and resolving problems.
I agree with the versioning, this would be useful for the conversion tool.

yuhu-:
outdated - closed

PR #183, opened 2025-11-30
FIX - Double conversion removed from Mqtt::handleWrite

yuhu-:


danielkucera:
can you please elaborate on what is going on here?

yuhu-:
I'm surprised that you want to know so much detail,...

The passed value was converted twice, which wasn't intended.
- first in Mqtt::handleWrite
- second in getVectorFromDouble

danielkucera:
I no longer have a good overview of the code so I am trying to get some insights.
Thanks.

PR #187, opened 2025-12-07
Updated UI

danielkucera:


danielkucera:
@yuhu- , @ajvdw , what do you think about this one?

preview: https://raw.githack.com/danielkucera/esp-arduino-ebus/interface/src/root.html

yuhu-:
This definitely looks better than the current version. It should be easy to expand if something is added or removed.

A few more thoughts:

* We should probably change load, save, and wipe to more common names like load, save, and delete.

* list and download deliver almost the same content; perhaps one command would suffice.

* scan, scan full, scan vendor, and participants – perhaps participants should be called scan results.

ajvdw:
> This definitely looks better than the current version. It should be easy to expand if something is added or removed.
> 
> A few more thoughts:
> 
> * We should probably change load, save, and wipe to more common names like load, save, and delete.
> * list and download deliver almost the same content; perhaps one command would suffice.
> * scan, scan full, scan vendor, and participants – perhaps participants should be called scan results.

Look good! I think we should limit the commands for the 'end user'. Most commands can be hidden because the effect of just clicking around can be destructive.

Another thought: 
- Brush Up  the UI of the config screen 
- add the 4 user commands to Upload Download Wipe and Store
- add link to the advanced commands.

PR #191, opened 2025-12-13
Added time synchronization and basic online monitoring.

yuhu-:
image

image

### DIFF
--- a/include/http.hpp
+++ b/include/http.hpp
@@ -7,7 +7,3 @@ void SetupHttpHandlers();
 
 extern WebServer configServer;
 extern IotWebConf iotWebConf;
-
-#if defined(EBUS_INTERNAL)
-void addLog(String entry);
-#endif
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -1,5 +1,6 @@
 #include "http.hpp"
 
+#include "log.hpp"
 #include "main.hpp"
 #include "mqttha.hpp"
 #include "schedule.hpp"
@@ -156,44 +157,7 @@ void handleResetStatistic() {
   configServer.send(200, "text/html", "Statistic reset");
 }
 
-#define MAX_LOG_ENTRIES 35
-String logBuffer[MAX_LOG_ENTRIES];
-int logIndex = 0;
-int logEntries = 0;
-
-String getTimestamp() {
-  time_t now = time(nullptr);
-  struct tm timeinfo;
-  localtime_r(&now, &timeinfo);
-
-  char buffer[30];
-  snprintf(buffer, sizeof(buffer), "%04d-%02d-%02d %02d:%02d:%02d.%03d",
-           timeinfo.tm_year + 1900, timeinfo.tm_mon + 1, timeinfo.tm_mday,
-           timeinfo.tm_hour, timeinfo.tm_min, timeinfo.tm_sec, millis() % 1000);
-
-  return String(buffer);
-}
-
-void addLog(String entry) {
-  String timestampedEntry = getTimestamp() + " " + entry;
-
-  logBuffer[logIndex] = timestampedEntry;
-
-  logIndex = (logIndex + 1) % MAX_LOG_ENTRIES;
-
-  if (logEntries < MAX_LOG_ENTRIES) logEntries++;
-}
-
-void handleLogData() {
-  String response = "";
-
-  for (int i = 0; i < logEntries; i++) {
-    int index = (logIndex - logEntries + i + MAX_LOG_ENTRIES) % MAX_LOG_ENTRIES;
-    response += logBuffer[index] + "\n";
-  }
-
-  configServer.send(200, "text/plain", response);
-}
+void handleLogData() { configServer.send(200, "text/plain", getLog()); }
 
 #endif
 
@@ -203,7 +167,7 @@ void handleRoot() {
     // -- Captive portal request were already served.
     return;
   }
-  
+
   extern const char root_html_start[] asm("_binary_static_root_html_start");
   configServer.send(200, "text/html", root_html_start);
 }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -8,6 +8,7 @@
 #if defined(EBUS_INTERNAL)
 #include <Ebus.h>
 
+#include "log.hpp"
 #include "mqtt.hpp"
 #include "mqttha.hpp"
 #include "schedule.hpp"
@@ -16,15 +17,14 @@
 #include "bus.hpp"
 #endif
 
-#include "client.hpp"
-#include "http.hpp"
-
 #include <ESPmDNS.h>
 #include <IotWebConfESP32HTTPUpdateServer.h>
 #include <esp_task_wdt.h>
 
+#include "client.hpp"
 #include "esp32c3/rom/rtc.h"
 #include "esp_sntp.h"
+#include "http.hpp"
 
 HTTPUpdateServer httpUpdater;
 
@@ -260,13 +260,9 @@ void wifiConnected() {
 #endif
 }
 
-void wdt_start() {
-  esp_task_wdt_init(6, true);
-}
+void wdt_start() { esp_task_wdt_init(6, true); }
 
-void wdt_feed() {
-  esp_task_wdt_reset();
-}
+void wdt_feed() { esp_task_wdt_reset(); }
 
 inline void disableTX() {
 #if defined(TX_DISABLE_PIN)
--- a/src/schedule.cpp
+++ b/src/schedule.cpp
@@ -4,6 +4,7 @@
 #include <set>
 
 #include "http.hpp"
+#include "log.hpp"
 #include "mqtt.hpp"
 #include "track.hpp"
 
--- a/include/log.hpp
+++ b/include/log.hpp
@@ -1,0 +1,6 @@
+#pragma once
+
+#include <Arduino.h>
+
+void addLog(String entry);
+String getLog();
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -1,0 +1,38 @@
+#include "log.hpp"
+
+#define MAX_LOG_ENTRIES 35
+String logBuffer[MAX_LOG_ENTRIES];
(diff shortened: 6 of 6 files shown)


src/http.cpp
```diff
@@ -153,6 +153,52 @@ void handleResetStatistic() {
...
+  return String(buffer);
+}
+
+void addLog(String entry) {
+  String timestampedEntry = getTimestamp() + " " + entry;
+
+  // If the buffer is full, shift all entries upwards
+  if (currentEntries >= MAX_LOG_ENTRIES) {
+    for (int i = 1; i < MAX_LOG_ENTRIES; i++)
+      logBuffer[i - 1] = logBuffer[i];  // Shift up
```
danielkucera: On each log it needs to move all the entries once we have MAX_LOG_ENTRIES reached.
Can you rather set
```
currentEntries = (currentEntries + 1) % MAX_LOG_ENTRIES
```
and then build the response with:
```
for (int i = currentEntries + 1; i < currentEntries + MAX_LOG_ENTRIES; i++)
...
response += logBuffer[i % MAX_LOG_ENTRIES] + "\n";
````

File: src/http.cpp [hunk 2]
```diff
@@ -161,51 +207,89 @@ void handleRoot() {
...
+  s += "<meta name=\"viewport\" content=\"width=device-width, initial-scale=1, user-scalable=no\"/>";
+  s += R"(
+    <script type='text/javascript'>
+      function confirmAction(action) {
+        return confirm("Are you sure you want to " + action + "?");
+      }
+    </script>
+  </head><body>
+  )";
+
```
danielkucera: Can we first merge #187 before this?

src/main.cpp
```diff
@@ -410,6 +474,24 @@ bool formValidator(iotwebconf::WebRequestWrapper* webRequestWrapper) {
     }
   }
 #if defined(EBUS_INTERNAL)
+  if (webRequestWrapper->arg(sntpServerParam.getId()).length() >
+      STRING_LEN - 1) {
+    String tmp = "max. ";
+    tmp += String(STRING_LEN);
```
danielkucera: Maximum length of DNS name is 255 characters.
Please define macro DNS_LEN and use where applicable.

src/main.cpp
```diff
@@ -787,13 +895,19 @@ void setup() {
   }
 
 #if defined(EBUS_INTERNAL)
+  if (sntpEnabledParam.isChecked()) {
+    initSNTP(sntpServer, sntpTimezone);
+    waitForSNTP();
```
danielkucera: What if there is an outage on that particular ntp server?
Waiting doesn't help anything here IMO.
From the logs it will be obvious if and when the NTP has been acquired.

src/schedule.cpp
```diff
@@ -730,25 +731,30 @@ void Schedule::handleEvents() {
   while (eventQueue.try_pop(event)) {
     if (event) {
       switch (event->type) {
-        case CallbackType::error:
+        case CallbackType::error: {
+          std::string payload = event->data.error + " : master '" +
+                                ebus::to_string(event->data.master) +
+                                "' slave '" +
+                                ebus::to_string(event->data.slave) + "'";
+          addLog(payload.c_str());
```
danielkucera, line 740: What about also sending to Mqtt?

include/http.hpp
```diff
@@ -7,3 +7,7 @@ void SetupHttpHandlers();
 
 extern WebServer configServer;
 extern IotWebConf iotWebConf;
+
+#if defined(EBUS_INTERNAL)
+void addLog(String entry);
```
danielkucera: I suggest creating a separate file with logging functions.
I feel like this will need to be extended in the future.

include/mqttha.hpp
```diff
@@ -11,6 +10,7 @@ class MqttHA {
  public:
   void setUniqueId(const std::string& id);
   void setRootTopic(const std::string& topic);
+  void setWillTopic(const std::string& topic);
```
danielkucera, line 13: Can you please move this last will into a separate PR?

README.md
```diff
@@ -339,10 +339,33 @@ Available MQTT commands.
...
 The following endpoints are not listed on the main page:
-- http://esp-ebus.local/api/v1/GetStatus
-- http://esp-ebus.local/api/v1/GetCounter
-- http://esp-ebus.local/api/v1/GetTiming
+|***endpoint***                            |***description***                                              
+|:-                                        |:-
+|http://esp-ebus.local/monitordata         |Raw value of monitor data     
+|http://esp-ebus.local/api/v1/GetStatus    |Displays the current status (JSON)
+|http://esp-ebus.local/api/v1/GetCounter   |Displays internal counters (JSON)
+|http://esp-ebus.local/api/v1/GetTiming    |Displays internal timings (JSON)
```
danielkucera: Please move unrelated doc changes to another PR or let's start using the Wiki feature. Do you have the rights to create stuff there?

src/http.cpp
```diff
@@ -227,6 +311,8 @@ void SetupHttpHandlers() {
   configServer.on("/api/v1/GetCounter", [] { handleGetCounter(); });
   configServer.on("/api/v1/GetTiming", [] { handleGetTiming(); });
   configServer.on("/reset", [] { handleResetStatistic(); });
+  configServer.on("/monitor", [] { handleMonitor(); });
+  configServer.on("/monitordata", [] { handleMonitorData(); });
```
danielkucera: I would call it simply log, not monitor

src/main.cpp
yuhu-: Do you really want to allow 255 characters for DNS ?

src/main.cpp
yuhu-: Maximum waiting time are 10 seconds at startup
```
void waitForSNTP() {
  const int timeout_limit = 100;  // Maximum wait time of 10 seconds
  const int delay_time = 100;     // Delay time in milliseconds
  ...
```

src/schedule.cpp
yuhu-: Error events are only sent to MQTT if `schedule.publishCounter` is set.

Or do you mean in general?

include/mqttha.hpp
yuhu-: This is already merged with  HA availability_topic field added #190

README.md
yuhu-: Would you like the HTTP documentation to be in the wiki? If so, I would also move the MQTT and Home Assistant documentation.

src/http.cpp
yuhu-: Initially I called it a log, but it's not a log because it's a ring buffer with a maximum of 35 entries.

src/http.cpp
yuhu-: Sure - do you have a timeline for the new UI?

include/http.hpp
yuhu-: Most of it is HTTP-related, which is why I placed it there. The ring buffer itself and the `addLog()` method could be located elsewhere.

src/main.cpp
danielkucera: I think we should be up to spec.

src/main.cpp
danielkucera: so what is the waiting good for?

include/mqttha.hpp
danielkucera: I noticed later, thanks.

README.md
danielkucera: Yes please. For now.

src/http.cpp
yuhu-: Changed

src/http.cpp
danielkucera: Let's please call it log.

src/main.cpp
yuhu-: Changed

src/main.cpp
yuhu-: Waiting for time is synchronized at startup once. Do you want a background task for that ?

danielkucera: But there is no action taken once the time is synchronized. Why do we need to keep checking at all?

src/http.cpp
yuhu-: Changed

README.md
yuhu-: Please add this until the Wiki is done

src/main.cpp
yuhu-: Reworked

README.md
yuhu-: Reverted - Wiki updated

src/http.cpp
```diff
@@ -161,51 +201,89 @@ void handleRoot() {
...
+                })
+                .then(data => { document.getElementById('logContainer').innerText = data; })
+                .catch(error => { console.error('Fetch error:', error); });
+        }
+        setInterval(fetchLogs, 500);  // Fetch logs every 500 milliseconds
+        fetchLogs();  // Initial fetch
+    </script>
+</body>
+</html>
+)rawliteral";
```
danielkucera: Please move as html file to static folder.

include/http.hpp
danielkucera: Please move to separate file. This will be extended in the future with multiple options so I want the codebase to be separated.

src/http.cpp
yuhu-: moved to static/log.html

README.md
danielkucera: Please check the headings levels in wiki. Thanks.

yuhu-: The wiki needed to be revised anyway. Next week...

src/http.cpp
danielkucera: I am thinking about the strategy for static html pages. Only longer ones in separate file or all?

yuhu-: I have no preferences.

include/http.hpp
yuhu-: log.hpp and log.cpp created

src/http.cpp
danielkucera: We will see later.

PR #194, opened 2025-12-14
Improved log page interface

danielkucera:
Disclaimer: Copilots work.

### DIFF
--- a/static/root.html
+++ b/static/root.html
@@ -59,6 +59,7 @@
                 <a href='/config' class="link-button">Configuration</a>
                 <a href='/firmware' class="link-button">Firmware Update</a>
                 <a href='/status' class="link-button">Status</a>
+                <a href='/log' class="link-button">Log</a>
                 <a href='/restart' onclick="return confirmAction('restart');" class="link-button danger">Restart</a>
             </div>
             <div class="info-box">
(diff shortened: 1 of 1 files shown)


yuhu-:
Should the "Log" button be generally available? (Currently only for the internal firmware)

It would fit well between the status and restart buttons.

yuhu-:
Ok, I tested.

- This version doesn't run very smoothly.
- The constant switching of the time and fetching is irritating.
- Downloads are limited to a maximum of 35 lines (ring buffer).
- Can the data be buffered in the browser (for more lines)?

yuhu-:
@ajvdw Could you please test this?

ajvdw:
It works, but I agree with Daniel. No need to keep it to 25 lines. Just start filling the screen with text. Just like 'List Commands'. I often use console.log for logging.....
There also the status service on port 5555. This could also be used as a logging service ....

danielkucera:
If the endpoint returned the data in some structured way like e.g. json array with micros(), timestamp, log, we could query it and only append new data and keep the old.
Now it only shows the exact content of query result.
Frequency can be adjusted too.

danielkucera:
Check now...

yuhu-:
I wanted to monitor bus messages directly in the browser without connecting to `ebusread`. Currently, the messages are displayed in reverse order (newest at the top), which might be convenient, but doesn't quite meet my needs. I realize this is a great feature for the average user and certainly helpful for PWM and bus ISR settings. Perhaps we could add category filters for bus and system messages, or allow users to select the output order.

image

yuhu-:
This version is definitely usable already. Perhaps you could move the Log button to the top between Status and Restart

PR #197, opened 2025-12-16
Commands page

danielkucera:
Copilot code.

### DIFF
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -16,15 +16,16 @@ void handleGetStatus() {
 }
 
 #if defined(EBUS_INTERNAL)
-void handleCommandsList() {
-  configServer.send(200, "application/json;charset=utf-8",
-                    store.getCommandsJson().c_str());
-}
-
 void handleCommandsPage() {
-  extern const char commands_html_start[] asm("_binary_static_commands_html_start");
+  extern const char commands_html_start[] asm(
+      "_binary_static_commands_html_start");
   configServer.send(200, "text/html", commands_html_start);
 }
+
+// void handleCommandsList() {
+//   configServer.send(200, "application/json;charset=utf-8",
+//                     store.getCommandsJson().c_str());
+// }
 
 void handleCommandsDownload() {
   String s = "{\"id\":\"insert\",\"commands\":";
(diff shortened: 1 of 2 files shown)


danielkucera:
https://raw.githack.com/danielkucera/esp-arduino-ebus/commands-page/static/commands.html

danielkucera:
I think the buttons deserve better names.

yuhu-:
- It works :-)
- Return Button to main menu? (This is also missing in the Log page)
- Clear Button ?
- Can you display the calculated size in bytes after loading, saving, and deleting?
- I had the saving problem again. After wiping, saving was possible.
- Do we really need the header in this form? Clearly in this way it is identical with mqtt command, but wouldn't it be easier without.
```
{
  "id": "insert",
  "commands": [
    {
```

danielkucera:
what about the button names?

yuhu-:
To be clear about their function: (order from left to right)
* Clear data - data in box will deleted
* Format data - data in box will formatted
* List / Show commands - current installed commands will be loaded into the box
* Insert / Update commands - installs the command from the box
* Upload - load a JSON file into the box and format it or install it directly ?
* Download - download the date from the box or the installed ?
* Load from NVS and install them
* Save installed commands into NVS
* Wipe or Delete saved commands in NVS

danielkucera:
Check now...

yuhu-:
nice

yuhu-:
the bytes seems to be odd...
26517 bytes (26494 chars) vs {"id":"load","status":"successful","bytes":6176}
image

danielkucera:
> the bytes seems to be odd...

How many UTF8 characters do you have there?

yuhu-:
The posted value is the return value of the serialize command

yuhu-:
- Clear button added
- Evalute button added / function added
- display loaded bytes from storage functions
- different color for apply buttons
- textare size fixed

image

Should I push?

yuhu-:
I will create a separate merge request.

danielkucera:
Feel free to push

yuhu-:
done - can't add you as reviewer and also can`t merge :-)

yuhu- (approved):

PR #202, opened 2025-12-21
Added command evaluation method; New http endpoint /commands/evaluate

yuhu-:


src/http.cpp
```diff
@@ -63,10 +63,61 @@ void handleCommandsDownload() {
...
 }
 
+void handleCommandsEvaluate() {
+  JsonDocument doc;
+  String body = configServer.arg("plain");
+
+  DeserializationError error = deserializeJson(doc, body);
+
+  if (error) {
+    configServer.send(403, "text/html", "INVALID JSON");
```
danielkucera: I am not a fan of CAPS for errors. It's obvious it's an error...

src/http.cpp
```diff
@@ -63,10 +63,61 @@ void handleCommandsDownload() {
...
+      for (JsonVariantConst command : commands) {
+        std::string evalError = store.evaluateCommand(command);
+        if (!evalError.empty()) {
+          configServer.send(403, "text/html", evalError.c_str());
+          return;
+        }
+      }
+      configServer.send(200, "text/html", "OK");
+    } else {
+      configServer.send(403, "text/html", "NO COMMANDS");
```
danielkucera: also here...

src/http.cpp
```diff
@@ -63,10 +63,61 @@ void handleCommandsDownload() {
...
+}
+
+void handleCommandsInsert() {
+  JsonDocument doc;
+  String body = configServer.arg("plain");
+
+  DeserializationError error = deserializeJson(doc, body);
+
+  if (error) {
+    configServer.send(403, "text/html", "INVALID JSON");
```
danielkucera: etc...

src/store.cpp
```diff
@@ -555,6 +597,55 @@ const std::string Store::getValuesJson() const {
...
+      (type == "bool" && !doc[field].is<bool>()) ||
+      (type == "int" && !doc[field].is<int>()) ||
+      (type == "float" && !doc[field].is<float>()) ||
+      (type == "uint8_t" && !doc[field].is<int>()) ||   // Cast to int for range
+      (type == "uint32_t" && !doc[field].is<int>()) ||  // Same for uint32_t
+      (type == "size_t" && !doc[field].is<int>()) ||    // Same for size_t
+      (type == "DataType" &&
+       (!doc[field].is<const char*>() ||
+        ebus::string_2_datatype(doc[field].as<const char*>()) ==
+            ebus::DataType::ERROR))) {
```
danielkucera: Can you please use enum instead of string type names?

Full copilot:
```
Yes — there are clearer and safer ways than comparing string names and repeatedly calling doc[field].is<...>(). Problems in the current approach:

    Using string literals for types is brittle and causes repeated string comparisons.
    Using is<int>() for unsigned types and size_t is imprecise (and doesn't check range).
    Presence check using !doc[field].is<JsonVariantConst>() is confusing — prefer containsKey or isNull().
    Special-case logic (DataType, ha_key_value_map) is mixed with generic checks.

Recommended approach

    Use a real enum for expected types instead of string names.
    Use JsonObjectConst::containsKey (or doc[field].isNull()) to test presence.
    For numeric types read as an integer (long) then validate range for uint8_t/size_t/uint32_t.
    Keep special handlers (DataType, object/map) separate in the switch.
    Return a descriptive error (or empty string) as you already do.

Minimal example (illustrative — adapt type names to your codebase):

    Define enum and FieldEvaluation: enum FieldType { FT_String, FT_Bool, FT_Int, FT_Float, FT_Uint8, FT_Uint32, FT_SizeT, FT_DataType, FT_HaKeyValueMap };

    struct FieldEvaluation { const char* name; bool required; FieldType type; };

    Implement a switch-based validator:

    const std::string Store::isFieldValid(const JsonDocument& doc, const std::string& field, bool required, FieldType type) { JsonObjectConst root = doc.as<JsonObjectConst>();

    if (required && !root.containsKey(field.c_str())) return "Missing required field: " + field;

    if (!root.containsKey(field.c_str()) || root[field.c_str()].isNull()) return ""; // not present and not required => ok

    JsonVariantConst v = root[field.c_str()];

    switch (type) { case FT_String: if (!v.is<const char*>()) return "Invalid type for field: " + field; break; case FT_Bool: if (!v.is<bool>()) return "Invalid type for field: " + field; break; case FT_Int: if (!v.is<long>()) return "Invalid type for field: " + field; break; case FT_Float: if (!v.is<float>() && !v.is<double>() && !v.is<long>()) return "Invalid type for field: " + field; break; case FT_Uint8: { if (!v.is<long>()) return "Invalid type for field: " + field; long val = v.as<long>(); if (val < 0 || val > 0xFF) return "Out of range for field: " + field; break; } case FT_Uint32: { if (!v.is<long>()) return "Invalid type for field: " + field; long val = v.as<long>(); if (val < 0) return "Negative value not allowed for field: " + field; // optionally check > UINT32_MAX break; } case FT_SizeT: { if (!v.is<long>()) return "Invalid type for field: " + field; long val = v.as<long>(); if (val < 0) return "Negative value not allowed for field: " + field; break; } case FT_DataType: if (!v.is<const char*>() || ebus::string_2_datatype(v.as<const char*>()) == ebus::DataType::ERROR) return "Invalid datatype for field: " + field; break; case FT_HaKeyValueMap: if (!v.is<JsonObjectConst>()) return "Invalid type for field: " + field; // call your isKeyValueMapValid on v.as<JsonObjectConst>() break; }

    return ""; }

    Then your fields array uses FieldType values (no string comparisons), and evaluateCommand loops the same way.

Other small notes

    Using root.containsKey(field) is clearer than doc[field].is<JsonVariantConst>() for existence.
    For numeric checks, prefer long/double depending on what ArduinoJson stores in your build; then range-check for unsigned types.
    Keep DataType and ha_key_value_map validation delegated to specialized checks as you already do.

If you want, I can convert your current isFieldValid + fields array to this enum-based implementation and update evaluateCommand and callers — want me to produce a patch?
```

yuhu-: JsonObjectConst::containsKey is deprecated - copilot is behind

https://arduinojson.org/v7/api/jsondocument/containskey/

danielkucera:
I was also thinking about one more thing...
For example read_cmd is string but it should only accept hexa chars. 

Can we introduce some regex matching, e.g. `^([0-9A-Fa-f]{2})+$`?

We could also do some logical checks, e.g.
If master is true then position should be more than length of command and such.

yuhu-:
about position
  bool master = false;                              // value of interest is in master or slave part
  size_t position = 1;                              // starting position within the data bytes, beginning with 1

danielkucera (approved):

PR #207, opened 2025-12-31
participant replaced by device

yuhu-:
- `Device` seems a more appropriate term.
- This is to prepare a device page with scan functions.

### DIFF
--- a/include/mqtt.hpp
+++ b/include/mqtt.hpp
@@ -27,30 +27,30 @@ struct IncomingAction {
       : type(IncomingActionType::Remove), command(), key(k) {}
 };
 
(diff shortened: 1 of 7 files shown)


danielkucera:
The spec is using the term `participant`. Why do you find `device` more appropriate?

yuhu-:
In my opinion and from a technical perspective, every eBUS participant is a device. Therefore, I prefer this term even though the specification uses "participant".

danielkucera:
For me participant implicitly means "someone else", device can be understood also as "myself" - the adapter.
But I am going to Ok this, I leave the decision on you.

PR #210, opened 2026-01-02
UI expanded and standardized

yuhu-:
- formatter used on html files
- status, values, device and statistic page added
- common.css and common.js added
- http endpoints moved
- fonts unified
- snake_case endpoints

@danielkucera & @ajvdw - can you check it out

I am thinking of removing the mqtt interface when the http endpoints are completed.

### DIFF
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -115,15 +115,20 @@ void handleCommandsRemove() {
     configServer.send(403, "text/html", "Json invalid");
   } else {
     JsonArrayConst keys = doc["keys"].as<JsonArrayConst>();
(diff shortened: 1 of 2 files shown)


danielkucera:
Very nice!
Since we have several static files now, can you please adjust the handler so it would be called in this manner?
```
configServer.on("/common.js", [] { handleStatic("application/javascript", asm("_binary_static_common_js_start")); });
```

danielkucera (approved):

yuhu-:
I've just added remove buttons to the commands page and made some minor improvements.

PR #212, opened 2026-01-06
Retrieve commands/values ​​ordered by key

yuhu-:
- do we need named arrays in json (commands / values) ?
for e.g. `{"values":[{..."`
- do we need the insert header in commands ?
for e.g. `{"id":"insert","commands":[{...`

I think we can do without it, what do you say?

### DIFF
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -40,16 +40,9 @@ void handleStatus() {
 
 #if defined(EBUS_INTERNAL)
 // commands
-void handleCommandsList() {
+void handleCommands() {
   configServer.send(200, "application/json;charset=utf-8",
                     store.getCommandsJson().c_str());
-}
-
-void handleCommandsDownload() {
-  String s = "{\"id\":\"insert\",\"commands\":";
-  s += store.getCommandsJson().c_str();
-  s += "}";
-  configServer.send(200, "application/json", s);
 }
 
 void handleCommandsEvaluate() {
@@ -61,7 +54,7 @@ void handleCommandsEvaluate() {
   if (error) {
     configServer.send(403, "text/html", "Json invalid");
   } else {
-    JsonArrayConst commands = doc["commands"].as<JsonArrayConst>();
+    JsonArrayConst commands = doc.as<JsonArrayConst>();
     if (!commands.isNull()) {
       for (JsonVariantConst command : commands) {
         std::string evalError = store.evaluateCommand(command);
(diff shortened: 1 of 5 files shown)


include/store.hpp
```diff
@@ -101,8 +101,9 @@ class Store {
   int64_t saveCommands() const;
   static int64_t wipeCommands();
 
-  static JsonDocument getCommandJson(const Command* command);
-  const JsonDocument getCommandsJsonDocument() const;
+  static JsonDocument getCommandJsonDoc(const Command* command);
```
danielkucera, line 104: I am wondering, wouldn't it make sense to convert Command to class and implement this as method?

src/http.cpp
```diff
@@ -49,8 +49,6 @@ void handleCommandsDownload() {
   String s = "{\"id\":\"insert\",\"commands\":";
   s += store.getCommandsJson().c_str();
   s += "}";
-  configServer.sendHeader("Content-Disposition",
-                          "attachment; filename=esp-ebus-commands.json");
```
danielkucera, line 53: Is there any problem with this? I like when it downloads automatically in browser.

include/store.hpp
yuhu-: One of the next steps is indeed to convert the command structure into a class. That's fine for now.

src/http.cpp
yuhu-: Are you using this API endpoint directly?

danielkucera:
> * do we need named arrays in json (commands / values) ?
>   for e.g. `{"values":[{..."`
> * do we need the insert header in commands ?
>   for e.g. `{"id":"insert","commands":[{...`

No, I would only return the array itself. Action and "content" (values/commands/...) can be determined by the endpoint URL and/or method.

But now I noticed we return values just as an array, there I would prefer key: value pairs.

yuhu-:
This commit changes this to an array like ...
```{"values":[{"key":"01","name":"Outside_Temperature","value":0.81,"unit":"°C","age":24},{"key":"02"```

yuhu-:
> > * do we need named arrays in json (commands / values) ?
> >   for e.g. `{"values":[{..."`
> > * do we need the insert header in commands ?
> >   for e.g. `{"id":"insert","commands":[{...`
> 
> No, I would only return the array itself. Action and "content" (values/commands/...) can be determined by the endpoint URL and/or method.
> 
> But now I noticed we return values just as an array, there I would prefer key: value pairs.

Then I will remove the array name and id...

src/http.cpp
danielkucera: Not now when we have the JS page but for a direct link it can come handy.

yuhu-: I just removed the download endpoint - the api becomes cleaner

PR #213, opened 2026-01-06
Use vertical text in commands table

danielkucera:
to maximize the amount of columns visible


static/common.css
yuhu-, line 1: since it's special for commandsTable and not common can you move it into the commands.html?

danielkucera: done

PR #218, opened 2026-01-07
Implement command write using http

danielkucera:


src/mqtt.cpp
```diff
@@ -280,6 +280,7 @@ void Mqtt::handleWrite(const JsonDocument& doc) {
       std::vector<uint8_t> writeCmd = command->write_cmd;
       writeCmd.insert(writeCmd.end(), valueBytes.begin(), valueBytes.end());
       schedule.handleWrite(writeCmd);
+      mqtt.publishResponse("write", "scheduled for key '" + key + "' name '" + command->name + "'");
```
yuhu-, line 283: Why do you need this line on MQTT?

danielkucera: this is unrelated but I noticed there was no response on write command from mqtt?

yuhu-: When the command is sent, a response is sent to MQTT.

Perhaps we could create a log entry when user interaction occurs, but it's ok for me here too

danielkucera: Well, the problem is that when the command is not sent (for whatever reason) there is no response.

And yes, we very much need better logging and error reporting.

Issue #219, opened 2026-01-07
readme


yuhu-:
**esp8266-arduino-ebus** maybe **esp-arduino-ebus**

danielkucera:
please check #220

PR #220, opened 2026-01-07
Improved readme

danielkucera:
Closes: #219


yuhu-:
Fewer symbols might make it seem less playful.

danielkucera:
what about now?

PR #222, opened 2026-01-08
values UI: add set/read functions

danielkucera:


src/http.cpp
```diff
@@ -189,7 +189,7 @@ void handleValuesWrite() {
         std::vector<uint8_t> writeCmd = command->write_cmd;
         writeCmd.insert(writeCmd.end(), valueBytes.begin(), valueBytes.end());
         schedule.handleWrite(writeCmd);
-        command->last = 0;  // force immediate update
+        //command->last = command->interval - 3;  // force immediate update
```
danielkucera: @yuhu- , what should we do about this?

yuhu-: just remove it then

yuhu-: this line updates the value for the HA - via http we do not need this in this way

yuhu- (approved):

PR #226, opened 2026-01-10
Alignment of MQTT topics with HTTP endpoints

yuhu-:
- class Track removed

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -530,23 +530,12 @@ char* status_string() {
                   reconnect_count);
   pos +=
       snprintf(status + pos, bufferSize - pos, "rssi: %d dBm\n", WiFi.RSSI());
-#if defined(EBUS_INTERNAL)
   pos +=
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -555,7 +542,7 @@ char* status_string() {
 
 #if defined(EBUS_INTERNAL)
   pos += snprintf(status + pos, bufferSize - pos, "loop_duration: %u us\r\n",
-                  loopDuration.value());
+                  loopDuration);
 #else
   pos += snprintf(status + pos, bufferSize - pos, "loop_duration: %u us\r\n",
                   loopDuration);
```
danielkucera: this seems to be the same?

yuhu-: It is the same, but without Track class

danielkucera: but why is it under the `if defined` condition when it is the same code for both cases?

yuhu-: You are right

danielkucera (approved):

PR #229, opened 2026-01-11
Remove command distance

### LINKED ISSUE #217: Error when trying write
command:
```
  {
    "key": "06",
    "name": "DHWTempSet",
    "read_cmd": "15b509030d4000",
    "write_cmd": "15b509050e40006801",
    "active": true,
    "interval": 60,
    "master": false,
    "position": 1,
    "datatype": "DATA2C",
    "divider": 1,
    "min": 1,
    "max": 100,
    "digits": 3,
    "unit": "°C",
    "ha": true,
    "ha_component": "sensor",
    "ha_device_class": "temperature",
    "ha_entity_category": "",
    "ha_mode": "auto",
    "ha_key_value_map": {},
    "ha_default_key": 0,
    "ha_payload_on": 1,
    "ha_payload_off": 0,
    "ha_state_class": "",
    "ha_step": 1
  },
```
request:
```
{"id":"write","key":"06","value":30}
```
response: nothing!

log:
```
10:38:21.363 INFO errorActiveMaster : master '' slave ''
```

### LINKED ISSUE #228: Schedule configuration - Command distance
In my opinion, this parameter is no longer useful or necessary. Since the interval is set individually for each command, the general setting is unnecessary.

I tested it with the value 0 and had no problems (the same error rate as with the value 1 = 1 second minimum waiting time between two active commands).

Perhaps we should allow the value 0 or remove this setting entirely.

EDIT: Since the device responds significantly better to external commands (MQTT or HTTP), I decided to remove it completely.

yuhu-:
- Changed `command distance` to `first command after start` between 5-60 seconds (10)
- Resolves #228 
- Resolves #217

### DIFF
--- a/include/schedule.hpp
+++ b/include/schedule.hpp
@@ -124,6 +124,8 @@ class Schedule {
 
   bool fullScan = false;
   uint8_t scanIndex = 0;
+  uint32_t distanceFullScans = 500;  // 500 milliseconds between 2 scans
+  uint32_t lastFullScan = 0;         // in milliseconds
 
   std::map<uint8_t, Device> allDevices;
 
--- a/src/schedule.cpp
+++ b/src/schedule.cpp
@@ -115,8 +115,7 @@ void Schedule::handleScan() {
       slaves.insert(slave.first);
 
   for (const uint8_t slave : slaves) {
-    std::vector<uint8_t> command;
-    command = {slave};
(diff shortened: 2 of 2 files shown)


yuhu-:
This affects the scanning capability.

yuhu-:
there is still a problem with full scan

PR #231, opened 2026-01-15
use single map for commands

danielkucera:


yuhu-:
This will increase the look up time for passive commands. On the other hand it will simplify the code. Did you test it?

danielkucera:
Is it not premature optimization? According to copilot:

Performance Analysis for ESP32-C3 with ~50 commands:

Current implementation ([findPassiveCommands](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)):

Algorithm: Linear scan of all 50 commands
Complexity: O(n) where n = ~50 commands
Per-query cost:
Loop through 50 map iterations: ~50 cycles
Each iteration: if (cmd->active) continue + ebus::contains(master, cmd->read_cmd) check
[ebus::contains()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is likely a binary search or substring match (~5-20 cycles for typical eBUS commands)
Estimated total: ~500-1000 CPU cycles per passive command query
Practical impact on ESP32-C3:

ESP32-C3 runs at 160 MHz (typical clock)
500-1000 cycles = 3-6 microseconds per query
Even with 10 passive queries per telegram: 30-60 microseconds
eBUS telegram processing typically takes hundreds of milliseconds, so this is negligible (<0.1% overhead)
Conclusion:
✅ No meaningful performance impact — the linear scan through 50 commands is trivial compared to eBUS bus latencies. You could safely have 200+ commands before this becomes a concern.

If you want to optimize further (future-proofing), you could rebuild the secondary index on read_cmd at insertion time, but it's not necessary now.

yuhu- (approved):
Just change it and we deal with it when ist becomes a problem

yuhu-:
Why do you want to change this anyway?

yuhu-:
I performed some performance measurements with 43 commands (35 active, 8 passive).
```
"Find":{"Last":38,"Mean":87,"StdDev":112,"Count":5009} - OLD
"Find":{"Last":27,"Mean":68,"StdDev":112,"Count":5005} - NEW
```
The modified implementation seems to be slightly better.

Measurements implementation
```
ebus::TimingStats findPassive;

std::vector<Command*> Store::findPassiveCommands(const std::vector<uint8_t>& master) {
  std::chrono::steady_clock::time_point t_start = std::chrono::steady_clock::now();

  std::vector<Command*> result;
  for (auto& kv : commands) {
    Command* cmd = &kv.second;
    if (cmd->active) continue;  // Skip active commands
    if (ebus::contains(master, cmd->read_cmd)) {
      result.push_back(cmd);
    }
  }

  std::chrono::steady_clock::time_point t_end = std::chrono::steady_clock::now();
  int64_t duration = std::chrono::duration_cast<std::chrono::microseconds>(t_end - t_start).count();

  findPassive.add(duration);
  return result;
}
```

danielkucera:
wow, that is surprising...

PR #235, opened 2026-01-17
Files renamed and split

### LINKED ISSUE #225: Conventions for naming classes and files
I intend to adjust the filenames to match the respective class names, if that's acceptable to you. The class names are already written in CamelCase.

yuhu-:
- Resolves #225 
- Logger class created
- ClientManager and ClientType classes moved to separate files

### DIFF
--- a/include/Logger.hpp
+++ b/include/Logger.hpp
@@ -2,13 +2,6 @@
 
 #include <Arduino.h>
 
-enum class LogLevel { DEBUG, INFO, WARN, ERROR };
-
-static const char* getLogLevelText(LogLevel logLevel) {
-  const char* values[] = {"DEBUG", "INFO", "WARN", "ERROR"};
-  return values[static_cast<int>(logLevel)];
-}
-
 // Simple circular buffer logger
 
 class Logger {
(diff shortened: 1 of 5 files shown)


include/Logger.hpp
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// Simple circular buffer logger
+
+class Logger {
+ public:
+  Logger(size_t maxEntries = 35);
+  ~Logger();
+
+  void add(LogLevel level, String message);
```
danielkucera: It is common to call the log function just `log()`

include/Logger.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+  String get();
```
danielkucera: This one I would call `getLogs()`

include/Logger.hpp
yuhu-: There are other common conventions too, but we could also use a library 
`logger.info(message)`
`logger.warn(message)`

danielkucera: nice!

PR #240, opened 2026-01-18
HA MQTT parallel publication reverted

### LINKED ISSUE #233: HA Discovery components not fully published - only ~3 of 8 sensors appear
## Firmware Version
v7.1-65-gce174e0-dirty

## Problem
When adapter has multiple commands configured with `ha: true`, only ~3 of 8 HA discovery components are published to MQTT. The remaining sensors never appear in Home Assistant.

The `/values` endpoint shows ALL sensors are being read correctly from eBUS, but only a subset get their HA discovery messages published to MQTT.

## Steps to Reproduce
1. Configure 8 commands via MQTT request (all with `ha: true`)
2. Restart adapter or trigger HA discovery
3. Only ~3 sensors appear in Home Assistant
4. Other sensors never receive discovery messages

## Root Cause
In `mqtt.cpp`, `checkOutgoingQueue()` only processes **1 message per 200ms**:

```cpp
void Mqtt::checkOutgoingQueue() {
  if (!outgoingQueue.empty() && millis() > lastOutgoing + outgoingInterval) {
    lastOutgoing = millis();
    OutgoingAction action = outgoingQueue.front();
    outgoingQueue.pop();
    // processes only 1 item
  }
}
```

With `outgoingInterval = 200` in `mqtt.hpp`, when `publishComponents()` enqueues 8+ HA components, it takes 1.6+ seconds to process all. This causes race conditions where only the first few components get published before timeouts or disconnects interrupt the sequence.

## Proposed Fix
1. Reduce `outgoingInterval` from 200ms to 50ms
2. Process multiple items per interval (e.g., up to 5)

```cpp
void Mqtt::checkOutgoingQueue() {
  if (outgoingQueue.empty() || millis() <= lastOutgoing + outgoingInterval) {
    return;
  }
  lastOutgoing = millis();

  int processed = 0;
  while (!outgoingQueue.empty() && processed < 5) {
    OutgoingAction action = outgoingQueue.front();
    outgoingQueue.pop();
    processed++;

    switch (action.type) {
      case OutgoingActionType::Command:
        publishCommand(action.command);
        break;
      case OutgoingActionType::Device:
        publishDevice(action.device);
        break;
      case OutgoingActionType::Component:
        mqttha.publishComponent(action.command, action.haRemove);
        break;
    }
  }
}
```

This allows all 8 components to publish in ~100ms instead of 1600ms.

## Environment
- Home Assistant 2024.x
- Boiler: Protherm Ray 18 KE
- eBUS commands: B511/01, B511/02, B511/00, B511/03

yuhu-:
- fixed #233

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -73,7 +73,7 @@ board_build.embed_txtfiles =
 
 lib_deps = 
     ${env.lib_deps}
-    https://github.com/yuhu-/ebus#75057ce
+    https://github.com/yuhu-/ebus#67f46e2
 
 [env:esp32-c3-internal-ota]
 extends = env:esp32-c3-internal
--- a/src/MqttHA.cpp
+++ b/src/MqttHA.cpp
@@ -312,8 +312,9 @@ MqttHA::Component MqttHA::createDiagnostic(const std::string& component,
 MqttHA::Component MqttHA::createDiagnosticUptime() const {
   Component c = createDiagnostic("sensor", "uptime", "Uptime");
   c.fields["state_topic"] = createStateTopic("state", "uptime");
-  c.fields["value_template"] =
-      "{{timedelta(seconds=((value|float)/1000)|int)}}";
+  c.fields["unit_of_measurement"] = "h";
+  c.fields["value_template"] = "{{((value|float)/1000/60/60)|round(2)}}";
+  c.fields["icon"] = "mdi:clock-outline";
 
   c.device["name"] = thingName;
   c.device["manufacturer"] = thingManufacturer;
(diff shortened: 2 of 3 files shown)


yuhu-:
Are there any doubts about that?

src/Mqtt.cpp
```diff
@@ -318,16 +318,10 @@ void Mqtt::checkIncomingQueue() {
...
 
 void Mqtt::checkOutgoingQueue() {
-  if (outgoingQueue.empty() || millis() <= lastOutgoing + outgoingInterval)
-    return;
-
-  lastOutgoing = millis();
-
-  int processed = 0;
-  while (!outgoingQueue.empty() && processed < 5) {
+  if (!outgoingQueue.empty() && millis() > lastOutgoing + outgoingInterval) {
```
danielkucera, line 321: The logic is too complicated IMO.
Can you just do in pseudocode:
- if queue empty - return
- if interval not passed - return
- do work

yuhu-: the same logic was before - just reverted - incoming is handled the same way

danielkucera (approved):

PR #247, opened 2026-01-24
Log entries added for the result of a bus request

yuhu-:


danielkucera:
What was wring with this one?

yuhu-:
There is nothing wrong but, I added the mutex to the logger and used this branch as base for the other merge request.

PR #254, opened 2026-01-25
Change struct command to class

yuhu-:


src/Command.cpp
```diff
@@ -0,0 +1,495 @@
...
+  for (const auto& field : fields) {
+    std::string error =
+        isFieldValid(doc, field.name, field.required, field.type);
+    if (!error.empty()) return error;  // Return the first error found
+  }
+
+  return "";  // No errors found
+}
+
+const std::string Command::isFieldValid(const JsonDocument& doc,
```
danielkucera, line 248: Can you use `std::optional` for the return value?

yuhu-: std::optimal requires C++17

danielkucera: ah, right...

PR #255, opened 2026-01-26
fix watchdog by adding main loop task

danielkucera:


danielkucera:
No, it will not solve, it might just restart in the case I had in this morning so we can find the cause.

PR #259, opened 2026-01-29
POC espressif esp-mqtt

yuhu-:
Size increased from
`RAM:   [==        ]  16.9% (used 55508 bytes from 327680 bytes)`
`Flash: [=======   ]  69.6% (used 1367426 bytes from 1966080 bytes)`
to
`RAM:   [==        ]  17.1% (used 55884 bytes from 327680 bytes)`
`Flash: [========  ]  75.8% (used 1490106 bytes from 1966080 bytes)`

Limitations:
The asynchronous transmission occurs every second with only one message.
[Message queued with esp_mqtt_client_enqueue() are delayed (IDFGH-14719) #15458](https://github.com/espressif/esp-idf/issues/15458)

Using esp_mqtt_client_publish instead of esp_mqtt_client_enqueue fixes the limitation and works smoothly.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -72,7 +72,7 @@ board_build.embed_txtfiles =
 
 lib_deps = 
     ${env.lib_deps}
-    https://github.com/yuhu-/ebus.git#bbc85d839b967c5d67da0ec78547e7f0299325c8
+    https://github.com/yuhu-/ebus.git#44a31f13716d161693ca2a61c04c40e90fc85129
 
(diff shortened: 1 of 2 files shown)


yuhu-:
I see several advantages to use this library, even though the increase in size is enormous.
- It's maintained and included in the framework.
- HTTP, WebSocket...and their secure variants.

danielkucera:
yes, I am pro

PR #260, opened 2026-01-30
POC PsychicMqttClient (based on espressif esp-mqtt)

yuhu-:
Size increased from
`RAM:   [==        ]  16.9% (used 55508 bytes from 327680 bytes)`
`Flash: [=======   ]  69.6% (used 1367426 bytes from 1966080 bytes)`
to
`RAM:   [==        ]  17.1% (used 55980 bytes from 327680 bytes)`
`Flash: [========  ]  76.1% (used 1495678 bytes from 1966080 bytes)`

Asynchronous transmission is limited. Instead, a message is transmitted immediately.

I would use esp-mqtt directly instead of another library.

### DIFF
--- a/include/Mqtt.hpp
+++ b/include/Mqtt.hpp
@@ -2,7 +2,7 @@
 
 #if defined(EBUS_INTERNAL)
 #include <ArduinoJson.h>
-#include <AsyncMqttClient.h>
+#include <PsychicMqttClient.h>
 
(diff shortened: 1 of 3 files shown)


danielkucera:
> I would use esp-mqtt directly instead of another library.

I agree.

yuhu-:
So I drop it

PR #261, opened 2026-01-30
Change struct device to class

yuhu-:
image

### DIFF
--- a/src/Schedule.cpp
+++ b/src/Schedule.cpp
@@ -439,8 +439,7 @@ const std::string Schedule::getDevicesJson() const {
   std::string payload;
   JsonDocument doc;
 
-  if (devices.size() > 0)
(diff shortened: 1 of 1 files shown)


src/Schedule.cpp
```diff
@@ -464,43 +435,12 @@ const std::string Schedule::getTimingJson() {
...
 const std::string Schedule::getDevicesJson() const {
   std::string payload;
   JsonDocument doc;
 
-  if (allDevices.size() > 0) {
-    for (const std::pair<uint8_t, Device>& device : allDevices)
-      doc.add(getDeviceJsonDoc(&device.second));
-  }
+  if (devices.size() > 0)
+    for (const auto& device : devices) doc.add(device.second.toJson());
```
danielkucera: is the if necessary?

yuhu-: right - I think it isn't necessary

PR #263, opened 2026-01-31
3 repetitions each for bus request and sending added

yuhu-:
Bus requests and transmissions are repeated up to a maximum of three times for each command, with the exception of full scan commands. This should help to resolve the issue of an active command not being sent.

### DIFF
--- a/src/Schedule.cpp
+++ b/src/Schedule.cpp
@@ -176,6 +176,9 @@ void Schedule::resetCounter() {
   seenMasters.clear();
   seenSlaves.clear();
 
+  busRequestFailed = 0;
+  sendingFailed = 0;
+
   if (ebusRequest) ebusRequest->resetCounter();
   if (ebusHandler) ebusHandler->resetCounter();
 }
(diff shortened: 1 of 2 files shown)


src/Schedule.cpp
```diff
@@ -522,11 +547,11 @@ void Schedule::handleEventQueue() {
...
   uint32_t currentMillis = millis();
 
-  // Check if scheduleCommand is stuck
-  if (scheduleCommand != nullptr && scheduleCommandSetTime > 0) {
-    if (currentMillis > scheduleCommandSetTime + scheduleCommandTimeout) {
-      scheduleCommand = nullptr;  // Clear the stuck command
-      scheduleCommandSetTime = 0;
+  // Check if activeCommand is stuck
+  if (activeCommand && activeCommand->setTime > 0) {
+    if (currentMillis > activeCommand->setTime + activeCommandTimeout) {
```
danielkucera: We should probably also check other millis comparisons.

GitHub Copilot Chat Assistant:

No — that comparison can break when millis() wraps. The safe, overflow-proof pattern is to subtract timestamps, not add them.

Why:

    millis() is (usually) a 32-bit unsigned value that wraps ~every 49.7 days.
    Doing currentMillis > setTime + timeout can fail when setTime + timeout overflows (wraps), producing a small value and making the comparison wrong.

Recommended change (use unsigned subtraction and >= for timeout semantics):

    Ensure the values are unsigned (uint32_t) and then: if ((uint32_t)(currentMillis - activeCommand->setTime) >= (uint32_t)activeCommandTimeout) { ... }

This handles wrap-around correctly because unsigned modulo arithmetic yields the correct elapsed time. Also keep your existing guard that setTime > 0 (so you don't treat an unset timestamp as elapsed).

PR #282, opened 2026-03-01
active buttons on main page

### LINKED ISSUE #281: The main page has no active buttons
Hello, I was using the old v7.1 well. Now I've tested Commit 07470b4 and latest Commit e6e5782 and on both have the same issue with the main page. I see it, but no one button is active. Can't to press them.

Image

But if I use the links: http://esp-ebus.local/config, http://esp-ebus.local/status, or http://esp-ebus.local/firmware, all is working well.
I've tested Chrome, Firefox on PC and mobile connected, but the same result.

yuhu-:
fix #281

### DIFF
--- a/static/root.html
+++ b/static/root.html
@@ -17,11 +17,11 @@
                 <a href='/config' class="link-button">Configuration</a>
                 <a href='/firmware' class="link-button">Firmware Update</a>
                 <a href='/status' class="link-button">Status</a>
-                <a href="/commands" class="link-button" id="Commands">Commands</a>
-                <a href="/values" class="link-button" id="Values">Values</a>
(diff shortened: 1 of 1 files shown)


static/root.html
```diff
@@ -44,17 +44,27 @@ <h1>esp-eBus Adapter</h1>
 
         function checkEndpoint() {
             const url = '/commands';
-            const links = document.querySelectorAll('.link-button');
+            const disabledIds = ['Commands', 'Values', 'Devices', 'Statistics', 'Logs'];
```
danielkucera: Can we add an additional class e.g. `internal` to each button, instead of listing them here?

PR #288, opened 2026-03-02
WiP: Implemented configuration manager class

danielkucera:


danielkucera:
Hi @yuhu- , did you have a chance to test it? Do you have some other notes to that?

yuhu-:
I assume after flashing this branch it starts with an empty configuration

danielkucera:
Yes, unfortunatelly. But the migration would be unreliable due to how many variants of eeprom contents there were over time...

yuhu-:
It worked. Bus ISR Window and Bus ISR Offset were empty – we could also offer default values ​​here (4300 and 70 µs).

The Wi-Fi connection is a bit choppy – perhaps some optimizations can be made here.

I would say merge it into the master.

yuhu-:
It looks like the adapter is connected to my Wifi and also offers the ebus-eBus AP

yuhu-:
After restarting, the adapter needs a few minutes to stabilize – possibly due to the simultaneous operation of the Wifi AP.

yuhu-:
I think, the WIFI isn't stable in this version. Maybe because the libraries are outdated ?

danielkucera:
iotweboconf is also using WiFi.h under the hood so it should not make any difference.
Eventually I want to switch to idf calls but in a separate PR, I want to finish this one first.
I've disabled AP after connecting. Can you please retest?

PR #302, opened 2026-03-09
Switch to esp_http_server

danielkucera:


danielkucera:
Yes, that will be next step.

PR #313, opened 2026-03-17
Rewrite client

danielkucera:


yuhu-:
And that works?

EDIT: It looks like you want to go back to writing directly on the bus. Right?

danielkucera:
no...

yuhu-:
What do you want to achieve?

danielkucera:
It doesn't work at all in current master.
I tried to vibe code it because I don't have deep understanding of how it works now but I failed.
My idea was so Client manager only handles new connections and ClientType has separate tx rx tasks.

yuhu-:
> My idea was so Client manager only handles new connections and ClientType has separate tx rx tasks.

That sounds like a good idea, because byte-by-byte reading and writing from the outside should occur with low latency. Especially for ebusd.

yuhu-:
I am currently revising the ebus library. This also includes the handling of external clients.

danielkucera:
Okay, I will wait then. If possible, make priority to the internal message sending. I was able to block it by external client so it stopped getting values. I would also appreciate a possibility to make requestBus() blocking with a timeout.

PR #341, opened 2026-04-26
Remove CMake project configuration

danielkucera:
Remove project configuration from CMakeLists.txt


yuhu- (approved):

danielkucera:
@yuhu- , the point here is that it fails the CI test.

Issue #351, opened 2026-05-22
ESP hangs when rebooting too quickly after saving commands.json (flash write race condition)


nicobresseleers:
Firmware: master‑2‑gd203480 (INTERNAL build)

Hardware: ESP32

Description:

When editing /commands and clicking Save, followed immediately (<1s) by Reboot, the device 
crashes and gui is gone..

Web UI becomes unresponsive, eBUS stops, and values remain at 0.
Only a power‑cycle restores operation.

Steps to reproduce:

Open /commands

Modify any entry

Click Save

Within 0.5 seconds click Reboot

Device hangs or boots with corrupted commands.json ?

Expected behavior:

Reboot should wait until flash write is complete

Or reboot button should be disabled during write

Or firmware should flush and sync before reboot

Actual behavior:

Flash write is interrupted

commands.json becomes partially written

Firmware hangs parsing corrupted JSON

Device becomes unresponsive

Root cause:

Flash write is asynchronous

No write‑lock

No flush

Reboot does not check write state

Proposed fix:

Add write‑lock around config save

Block reboot while writing

Flush file before reboot

Disable reboot button for 1 second after Save

Not 100% that's what happen, but it happend twice... (in that area of writing commands and rebooting)

ludrao:
I encountered a different command-save crash while investigating INTERNAL mode, and although it had another root cause, I also changed the storage code in a way that may help with this issue.

I have not reproduced the exact "Save then Reboot within 0.5 s" case described here, so consider this only a robustness suggestion.

Currently, writing directly to commands.json means that the previously valid file can be truncated before the new contents are fully written.

I changed the save sequence to:

write commands.json.tmp
        ↓
fflush()
        ↓
fclose()
        ↓
rename commands.json -> commands.json.bak
        ↓
rename commands.json.tmp -> commands.json
        ↓
remove commands.json.bak

If writing the temporary file fails, the existing commands.json is never touched.

Simplified idea:

```
FILE* file = fopen("/littlefs/commands.json.tmp", "wb");
// write complete JSON
fflush(file);
fclose(file);

rename(commandsPath, backupPath);

if (rename(tempPath, commandsPath) != 0) {
    rename(backupPath, commandsPath);
    return -1;
}

remove(backupPath);
```

This also proved useful during my testing: when a save operation failed for another reason, the previously saved 28-command configuration remained intact and was loaded again after reboot instead of leaving an empty/corrupted command file.

[...]

Issue #335, opened 2026-04-12
Latest V7.2-internal firmware always restarts when storing commands


therealdt:
Storing commands using the web-interface always fails resetting the device.

Firmware
Version: v7.2
SDK: 5.5.3
Adapter_HW_Version: pre-7.0
Adapter_HW_Version_Raw: 0
Clock_Speed: 160
Apb_Speed: 80000000
Chip
Chip_Revision: 4
Flash_Chip_Size: 4194304

I have flashed the firmware using: esptool.py write_flash 0x0 firmware-fullflash-HW_v5.x-internal-185833.bin

Inserting the same commands using the API /api/v1/commands/insert seems to work. However, storing using the API "/api/v1/commands/save" is not possible - response: "Not found".

danielkucera:
`/api/v1/commands/save` is a POST endpoint.
Are you using POST method? Can you show how you are calling this?

danielkucera:
Resetting might be related to https://github.com/yuhu-/ebus/issues/14
An update is coming soon.

therealdt:
You are right, however the POST-method also usually doesn't work - I always get: 
curl: (56) Recv failure: Connection reset by peer

And the devices seems to restart and the inserted commands are lost.

danielkucera:
do you have a chance to capture output from USB serial port when connected to a PC and mentioned post is executed?

therealdt:
This is the serial port log:


```
W (160141) wifi:m f null
 
 
abort() was called at PC 0x42112c2f on core 0
Core  0 register dump:

MEPC    : 0x40386f58  RA      : 0x40386f16  SP      : 0x3fcc3fe0  GP      : 0x3fc96800  
TP      : 0x3fcc4360  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130  
S0/FP   : 0x00000004  S1      : 0x3fcc4044  A0      : 0x3fcc400c  A1      : 0x3fcc4042  
A2      : 0x00000000  A3      : 0x3fcc4039  A4      : 0x00000001  A5      : 0x3fca1000  
A6      : 0x7a797877  A7      : 0x76757473  S2      : 0x50001cd0  S3      : 0x3fcd9e64  
S4      : 0x3fcde814  S5      : 0x3fcd9f84  S6      : 0x3fcc4148  S7      : 0x00000000  
S8      : 0x00000000  S9      : 0x00000000  S10     : 0x00000000  S11     : 0x00000000  
[...]

PR #340, opened 2026-04-24
transition to new library structure

yuhu-:
**HTTP enpoints**
- [x] Configuration 
- [x] Firmware upgrade
- [x] Status
- [x] ADC
- [x] Commands - streaming for full upload of big json files
- [x] Cron - untested
- [x] Values
- [x] Devices
- [x] Metrics
- [x] Logs

**MQTT**
- [x] Home Assistant - support new configuration structure
- [x] fix Home Assistant autodiscovery for ha number_temperature

**OTA**
- [x] not working - ESPOTA failure: recv failed errno=104

**External Client**
- [x] ebusread works smoothly

**fixed findings**
- [x] timing of first bus attempt
- [x] test with different hardware 
  - tested with 7.x and 
  - pre-7.x - knowing issued: ADC isn't working

**TODO's**
- [x] move to #362 

**Key changes**
- **env:esp32-c3-internal-simulation profile** - Testing/Development without an eBUS connection
- **catch tests** - tests for Command and CommandManager in test_host/
- **stack first** - almost everything is now created on stack instead of heap
- **naming convention** - applied
- **ESP WIFI static buffer instead of dynamic** was the game changer to fix random crashes
  - CONFIG_ESP_WIFI_STATIC_TX_BUFFER=y
  - CONFIG_ESP_WIFI_STATIC_RX_BUFFER_NUM=6
  - CONFIG_ESP_WIFI_DYNAMIC_RX_BUFFER_NUM=6
  - CONFIG_ESP_WIFI_STATIC_TX_BUFFER_NUM=4
  - CONFIG_ESP_WIFI_MGMT_SBUF_NUM=16
- **Modified command structure to accommodate up to 4 data fields** (compiler macro)
    -DCOMMAND_CAPACITY=96                   # Maximum number of read / passive commands to store
    -DCOMMAND_MAX_FIELDS=4                # Max fields per command
    -DWRITE_CMD_CAPACITY=16                # Max commands with write_cmd (separate storage)
- **Command definition redesigned to work with profiles**
  - profile definition are in json format and get converted at compile time into header files
  - **built in profiles:** data_profiles.json / ha_profiles.json
  - **user profiles:** data_profiles_user.json / ha_profiles_user.json
  - **scripts:** scripts/generate_data_profiles.py / scripts/generate_ha_profiles.py
  - **header:**  include/data_profile_gen.hpp / include/ha_profile_gen.hpp

Example command definition with profiles:
```
  {
    "key": "01",
[...]

Issue #332, opened 2026-04-11
Locked up


Th3M1k3y:
I just made it completely lock up, after saving this
```
[
  {
    "key": "04",
    "name": "storageTemp",
    "read_cmd": "08b509030d1700",
    "write_cmd": "0000",
    "active": true,
    "interval": 30,
    "master": false,
    "position": 1,
    "datatype": "DATA2C",
    "divider": 1,
    "min": 1,
    "max": 100,
    "digits": 1,
    "unit": "°C",
    "ha": true,
    "ha_component": "sensor",
    "ha_device_class": "temperature",
    "ha_entity_category": "",
    "ha_mode": "auto",
    "ha_key_value_map": {},
    "ha_default_key": 0,
    "ha_payload_on": 1,
    "ha_payload_off": 0,
    "ha_state_class": "",
    "ha_step": 1
  },
  {
    "key": "05",
    "name": "WaterPressure",
    "read_cmd": "08b509030d0200",
    "write_cmd": "0000",
    "active": true,
    "interval": 30,
    "master": false,
    "position": 1,
    "datatype": "DATA2C",
    "divider": 1,
    "min": 1,
    "max": 100,
    "digits": 1,
    "unit": "",
    "ha": true,
    "ha_component": "sensor",
    "ha_device_class": "pressure",
    "ha_entity_category": "",
    "ha_mode": "auto",
    "ha_key_value_map": {},
    "ha_default_key": 0,
    "ha_payload_on": 1,
    "ha_payload_off": 0,
    "ha_state_class": "",
    "ha_step": 1
  },
  {
    "key": "06",
    "name": "Gasvalve3UC",
    "read_cmd": "08b509030dDB00",
    "write_cmd": "0000",
    "active": true,
    "interval": 30,
    "master": false,
    "position": 1,
    "datatype": "DATA2C",
    "divider": 1,
    "min": 1,
    "max": 100,
    "digits": 1,
    "unit": "",
    "ha": true,
    "ha_component": "binary_sensor",
    "ha_device_class": "running",
    "ha_entity_category": "",
    "ha_mode": "auto",
    "ha_key_value_map": {},
    "ha_default_key": 0,
    "ha_payload_on": 15,
    "ha_payload_off": 240,
    "ha_state_class": "",
    "ha_step": 1
  }
]
```
The webserver stopped responding, the led on the microcontroller was still on, one led on in the corner of the pcb, the other flashing.

[...]

PR #360, opened 2026-07-30
fix(deps): pin patched ebus fork to fix crash on /api/v1/commands/save

macraj:
## Summary

Fixes the reproducible crash reported in #335 and #351: `POST /api/v1/commands/save`
aborts the device and reboots it, wiping the entire command list (not just the
newly-saved entries).

## Root cause

`esp32-c3-internal` pins `yuhu-/ebus` at commit `14869188c` (2026-03-16). At that
commit, `BusFreeRtos::s_onFallingEdge`/`s_onBusIsrTimer` (the GPIO/timer ISR
trampolines) are correctly marked `IRAM_ATTR`, and the GPIO ISR service is
installed with `ESP_INTR_FLAG_IRAM` — but the actual instance methods they
immediately call, `onFallingEdge()` and `onBusIsrTimer()`, are **not** marked
`IRAM_ATTR`. Since these are separate (non-inlined) functions, the CPU ends up
needing to fetch flash-resident instructions from inside an ISR. If that ISR
fires while flash cache happens to be disabled — which is exactly what happens
during any flash write, including `Store::saveCommands()` — the result is a
"Cache error: Cached memory region accessed while ibus or cache is disabled"
panic. This matches the symptom and diagnosis in yuhu-/ebus#14 ("Interrupt
handlers causing crash"), and explains why it's intermittent/timing-dependent
(only crashes if an eBUS interrupt happens to land inside the cache-disabled
window) — matching #351's "flash write race condition" title.

Confirmed this same gap is already fixed on `yuhu-/ebus` current `master}`
(`onFallingEdge`/`onBusIsrTimer` are `IRAM_ATTR` in the refactored
`bus_esp.cpp`) — but that library has since undergone a full namespace/directory
refactor incompatible with this project's current integration code, so bumping
straight to `master` isn't a safe drop-in here.

## Fix

Forked `yuhu-/ebus` at the exact same pinned commit
(`macraj/ebus@4deff5b`, branched from `14869188c`) and applied the minimal
`IRAM_ATTR` fix on top — same base, only the ISR-safety attributes added:
- `BusFreeRtos::onFallingEdge()` / `onBusIsrTimer()` marked `IRAM_ATTR`
  (declaration + definition)
- `Request::busRequestAddress()` (called from `onBusIsrTimer()`) inlined in
  the header instead of left as a separate non-IRAM translation unit, since
  `Request.hpp` is shared/portable code without ESP-IDF headers available for
  `IRAM_ATTR` directly

This PR just repoints `platformio.ini`'s `esp32-c3-internal` lib_deps at that
patched commit. No other changes.

## Test plan
- [x] `pio run -e esp32-c3-internal` builds cleanly against the patched fork
      (Flash 85.7%, RAM 14.5% — effectively identical to the unpatched build)
- [x] Verified the fetched dependency in `.pio/libdeps` actually contains the
      `IRAM_ATTR` additions before building
- [ ] Not yet verified against a live device reboot-under-`commands/save` repro
      (no serial/JTAG access on hand) — would appreciate someone who can
      reproduce #335 confirming this resolves it

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -59,9 +59,9 @@ build_flags =
     -DPWM_PIN=6
     -DEBUS_INTERNAL=1
 
-lib_deps = 
+lib_deps =
     ${env.lib_deps}
-    https://github.com/yuhu-/ebus.git#14869188cce168d96892d92fb5471cd4342e69a2
+    https://github.com/macraj/ebus.git#4deff5b5274306e35574fe2169e8a83fc82af4be
 
 [env:esp32-c3-internal-ota]
 extends = env:esp32-c3-internal


macraj:
Update: verified live on real hardware (a device that reliably crashed on the very first `commands/save` call before this fix).

After flashing with this patch applied: 6 consecutive `insert` + `save` cycles, zero reboots (uptime climbed continuously throughout, no resets). Then triggered a real restart (`GET /restart`) - all 12 saved commands loaded correctly from flash on boot, confirming persistence survives a real power cycle, not just repeated writes in a single session.

Build/release with both fixes applied: https://github.com/macraj/esp-arduino-ebus/releases/tag/v7.2.2

danielkucera:
@yuhu- , please have a look

yuhu-:
The first part looks like your PR https://github.com/yuhu-/ebus/issues/14 - if it works why not

EDIT: I compared this to the ebus master branch, and I've already done that there as well.

EDIT2: You already have half done with your own branch https://github.com/yuhu-/ebus/compare/master...danielkucera:ebus:isr-fix - I would suggest to added the missing pieces there and use your own PR https://github.com/danielkucera/esp-arduino-ebus/pull/348

Gandalf-the-Grey:
Confirming macraj's IRAM fix (#360) on **hardware rev 7.0** (earlier reports were v6.3).

On stock v7.2-internal with a live Vaillant eBUS, storing commands (`POST /api/v1/commands/save`) reliably crashed/rebooted the device and wiped the command list (matches #335 / #332 / #351). I built the firmware from source with the ISR `IRAM_ATTR` fix applied and OTA-flashed it. With the bus active, `/save` now completes cleanly ("N bytes saved") with zero reboots across repeated saves, and the saved commands reload from flash after a restart and a full power-cycle.

So the fix resolves the save-crash on HW 7.0 as well. Thanks for tracking it down!

danielkucera:
Hi guys, would you be willing to test #340 instead?

yuhu-:
this may be outdated

danielkucera:
@macraj , please test latest master. I believe this is obsolete now.

PR #365, opened 2026-09-06
Modularize HTTP API

yuhu-:
resolve issue #364

partly support issue #366


src/api/commands_api.cpp
```diff
@@ -0,0 +1,324 @@
...
+  });
+  httpd_resp_send_chunk(req, nullptr, 0);
+  return ESP_OK;
+}
+
+esp_err_t CommandsApi::handleCommandsEvaluate(httpd_req_t* req) {
+  HttpUtils::StreamingReader sr(req);
+  if (!sr.isValid() || !sr.feedAll()) {
+    HttpUtils::sendErrorResponse(req, "400 Bad Request", "evaluate",
+                                 "Request body too large or invalid");
```
danielkucera: please split the cases for invalid and too large

yuhu-: done

yuhu-:
ready to review

danielkucera:
can we instead of calling handlers setup from http.c , call it from inside of the modules during init? That way we could disable specific modules using just a define inside of that module file.

yuhu-:
Do you mean registerHandlers?

Your proposal requires a restructuring of the respective classes. We should start with the current separation.

danielkucera (approved):

